PR #5, opened 2024-09-19
Update publish command with new components

bensherred:
Adds the new free components to the publish command and README


calebporzio:
Agh, thanks ben, I'll try to be aware of that publish command going forward

PR #6, opened 2024-09-20
add tailwind RTL support

3adeling:
I added tailwind RTL support. below list of modifications

1. replaced `ml`, `mr`, `pl`, `pr`, `ml-auto` with `ms`, `me`, `ps`, `pe`, `ms-auto`
2. replaced `rounded-r-none`, `rounded-l-none` with `rounded-e-none`, `rounded-s-none`
3. replaced `border-r`, `border-r-0` with `border-e`, `border-e-0`
4. rotate `chevron-right` icon with 180 degrees in case of RTL.

### DIFF
--- a/stubs/resources/views/flux/button/group.blade.php
+++ b/stubs/resources/views/flux/button/group.blade.php
@@ -3,8 +3,8 @@ $classes = Flux::classes()
     ->add('flex group/button')
     ->add([ // Make the first, middle, and last buttons have proper border radiuses...
         '[&>*:not(:last-child):not(:first-child)]:rounded-none',
-        'first:*:rounded-r-none',
-        'last:*:rounded-l-none',
+        'first:*:rounded-e-none',
+        'last:*:rounded-s-none',
     ])
     ;
 @endphp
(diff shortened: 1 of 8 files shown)


3adeling:
@calebporzio RTL is easy to achive using tailwindcss. they added the ability to use start and end instead of right and left. so whenever you are using something like `left-*` or `right-*` always look if tailwind have other variants for `start-*` and `end-*`.

for other cases like `space-x-*` for RTL you will need to add `rtl:space-x-reverse`.

I can't remember any other cases.

calebporzio:
Yeah, I forget why I didn't do that - maybe I started and just forgot to continue with it? Anyhow, too close to launch to merge something like this. AND I'd have to do this for the whole private flux-pro repo. So going to close it now but thanks anyways.

pau1phi11ips:
This is quite a big change. Bootstrap did exactly the same between v4 and v5 and it made upgrading so much harder than it needed to be.

I still prefer to use left/right instead of start/end, if this could be an option.

3adeling:
@pau1phi11ips start and left are already released by tailwind. if @calebporzio don't want to use start/end instead of right/left it will require additional work to make all mentioned classes `ltr:*` and then add the reverse `rtl:*`.

ex. 

1. `ltr:left-0 rtl:right-0`
2. `ltr:pr-4 rtl:pl-4` 
3. `ltr:rounded-r-none rtl:rounded-l-none`

the only one doesn't require adding `ltr:*` is `space-x-*`, it only requires adding `rtl:space-x-reverse`

PR #26, opened 2024-09-24
Fix link current check to strip the app url

### LINKED ISSUE #23: navlist.item active only works with direct links and not routes
I'm not sure if this is a feature request or a bug but if you specify a route instead of a direct link in a navbar item, the navbar item will not become active.   This is documented [here](https://fluxui.dev/components/navbar#detecting-the-current-page), but doesn't specifically mention that routes cannot be used.

Does not show as active:
```php
<flux:navlist.item href="{{ route('news') }}">Latest News</flux:navlist.item>
```

Shows as active:
```php
<flux:navlist.item href="/news">Latest News</flux:navlist.item>
```

joshhanley:
Currently the `button-or-link` component doesn't recognise URLs generated using the `route()` helper as being the current route.

So this won't have the current page highlighting:

```blade
<flux:navlist.item href="{{ route('news') }}">Latest News</flux:navlist.item>
```

This PR fixes that.

Fixes #23

### DIFF
--- a/stubs/resources/views/flux/button-or-link.blade.php
+++ b/stubs/resources/views/flux/button-or-link.blade.php
@@ -5,6 +5,8 @@
 ])
 
 @php
+$href = $href ? (string) str($href)->after(config("app.url"))->rtrim("/")->finish('/') : $href;
+
 $current = $current === null ? ($href ? request()->is($href === '/' ? '/' : trim($href, '/')) : false) : $current;
 @endphp
 
--- a/stubs/resources/views/flux/button-or-link.blade.php
+++ b/stubs/resources/views/flux/button-or-link.blade.php
@@ -5,7 +5,7 @@
 ])
 
 @php
-$href = $href ? (string) str($href)->after(config("app.url"))->rtrim("/")->finish('/') : $href;
+$href = $href ? (string) str($href)->after(config('app.url'))->rtrim('/')->finish('/') : $href;
 
 $current = $current === null ? ($href ? request()->is($href === '/' ? '/' : trim($href, '/')) : false) : $current;
 @endphp


stubs/resources/views/flux/button-or-link.blade.php
```diff
@@ -5,6 +5,8 @@
 ])
 
 @php
+$href = $href ? (string) str($href)->after(config("app.url"))->rtrim("/")->finish('/') : $href;
```
bensherred, line 8:
```suggestion
$href = $href ? (string) str($href)->after(config('app.url'))->rtrim("/")->finish('/') : $href;
```

bensherred (approved):

trimline-gaiustemple:
Doesn't seem to be fixed for me on v1.0.1

Doesn't work
`<flux:navlist.item icon="chart-bar-square" href="{{ route('dashboard.index') }}">Overview</flux:navlist.item>`

Works
`<flux:navlist.item icon="chart-bar-square" href="{{ route('dashboard.index') }}" :current="request()->routeIs('dashboard.index')">Overview</flux:navlist.item>`

SRWieZ:
After this PR, I have a trailing "/" on all my links. This is a weird thing to do and can cause issues with some web servers.

Example of the link that it generates:  
`https://myproject.com/some/route?parameter=value/`

trimline-gaiustemple:
@joshhanley  can `->finish('/')` be removed from line 8 [button-or-link.blade.php](https://github.com/livewire/flux/pull/26/files#diff-f1881990f6a3a4ad7da4ff3930296fe002d93d6cdbcb86134e53ec9f193f0eeb)?

PR #35, opened 2024-09-24
Ensure button and link button display the same

### LINKED ISSUE #28: Button gets full width with href
When you use a href attribute on the buttons it becomes an <a> element and gets a fullwith property.
If you remove the href it is just a button and it is styled correctly.

joshhanley:
Currently if you have a standard button and a link button, the standard button will be small but the link button will expand to fill all available space (see first screenshot below).

```blade
<flux:button>Test</flux:button>
<flux:button href="/">Test</flux:button>
```

**Before - button inline-block, link acting like a block**
image

The reason for this is buttons are `display:inline-block` by default where as links are `display:block`.

So instead of using `flex` on the button component, we should use `inline-flex`. This PR makes that change (see second screenshot below).

**After - both acting like inline block**
image

Fixes #28.

### DIFF
--- a/stubs/resources/views/flux/button/index.blade.php
+++ b/stubs/resources/views/flux/button/index.blade.php
@@ -23,7 +23,7 @@ $iconVariant ??= ($size === 'xs')
     : ($square ? 'mini' : 'micro');
 
 $classes = Flux::classes()
-    ->add('flex items-center font-medium justify-center gap-2 whitespace-nowrap')
+    ->add('inline-flex items-center font-medium justify-center gap-2 whitespace-nowrap')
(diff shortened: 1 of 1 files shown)


bensherred (approved):
Thanks @joshhanley

giacomovernoni:
The button still renders as full width if the inset option is used.
```
<div class="mt-12">
    <p class="mb-3">
        Some text
    </p>
    <flux:button variant="ghost" inset>Button with href</flux:button>
</div>

<div class="mt-12">
    <p class="mb-3">
        Some text
    </p>
    <flux:button variant="ghost" inset href="#">Button with href and inset</flux:button>
</div>
```

flux button inset

PR #45, opened 2024-09-24
Fix TTY mode on windows

bensherred:
This fixes the activate command caused by TTY mode not being supported on Windows platforms.


calebporzio:
awesome, thanks Ben!

PR #77, opened 2024-09-25
Fix for tabs not being published 

idealerror:
Tabs stubs folder is called 'tab' and not 'tabs'

### DIFF
--- a/src/Console/PublishCommand.php
+++ b/src/Console/PublishCommand.php
@@ -48,7 +48,7 @@ class PublishCommand extends Command
             'Select' => ['select', 'option', 'options'],
             'Switch' => ['switch'],
             'Table' => ['table', 'columns', 'column', 'rows', 'row', 'cell', 'pagination'],
-            'Tabs' => ['tab'],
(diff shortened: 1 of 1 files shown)


src/Console/PublishCommand.php
```diff
@@ -48,7 +48,7 @@ class PublishCommand extends Command
             'Select' => ['select', 'option', 'options'],
             'Switch' => ['switch'],
             'Table' => ['table', 'columns', 'column', 'rows', 'row', 'cell', 'pagination'],
-            'Tabs' => ['tabs'],
+            'Tabs' => ['tab'],
```
bensherred, line 51: Hey @idealerror, thanks for the PR. You will need to include both `tabs` and `tab` due to there being a `tabs` component in the root of the stubs folder:

```suggestion
            'Tabs' => ['tab', 'tabs'],
```

idealerror: Fixed. I didn't see a folder for tabs, missed the blade component.

bensherred (approved):
Thanks @idealerror

Issue #67, opened 2024-09-25
Button and other Components adding slash to href


daerentis:
The button from the docs (https://fluxui.dev/components/button) is adding a slash to the end of the ``href``.
This breaks the ``foo`` parameter on the page that this link is directing to.

This:

```html
<flux:button href="https://www.example.com?foo=bar">
    Visit Me
</flux:button>
```

Generates this:

```html
<a href="https://www.example.com/?foo=bar/" class="inline-flex items-center font-medium justify-center gap-2 whitespace-nowrap disabled:opacity-50 dark:disabled:opacity-75 disabled:cursor-default disabled:pointer-events-none h-10 text-sm rounded-lg px-4  bg-white hover:bg-zinc-50 dark:bg-zinc-700 dark:hover:bg-zinc-600/75 text-zinc-800 dark:text-white border border-zinc-200 hover:border-zinc-200 border-b-zinc-300/80 dark:border-zinc-600 dark:hover:border-zinc-600 shadow-sm group-[]/button:-ml-[1px] group-[]/button:first:ml-0" data-flux-button="data-flux-button">
    Visit Me
</a>
```

joshhanley:
@daerentis thanks for reporting! I have submitted a PR #103 with a fix.

PR #97, opened 2024-09-26
Remove trailing '/' from button-or-link.blade.php

gaiustemple:
Related to issue #67 - a trailing slash is getting added to links since pr #26 , which causes unexpected behaviour when query parameters are being set.


joshhanley:
@gaiustemple thanks for the PR! I've submitted a PR with a simpler implementation which might be easier to maintain, see PR #103

gaiustemple:
@joshhanley looks great, will close this one

Issue #121, opened 2024-09-27
flux:menu.item extra double quote


codeldev:
In component flux:menu.item on Line 45

There is an additional double quote that should be removed.

<img width="623" alt="Screenshot 2024-09-27 at 07 35 39" src="https://github.com/user-attachments/assets/b9f2b05d-ab2b-4cc3-8f5c-bace141e1f1a">

bensherred:
This has now been fixed and will be available in the next release

PR #122, opened 2024-09-27
Fix extra double quote in flux.menu.item component #121

andreilungeanu:
Removed an extra double quote in the flux:menu.item component on Line 45. 
This resolves issue #121 , which was causing syntax errors in the component.


bensherred (approved):
Thanks @andreilungeanu

PR #178, opened 2024-09-30
Add loading indicator for Livewire buttons

calebporzio:
This PR detects a "Livewire button" by the presence of `wire:click="..."` or `type="submit"` on a `<flux:button>` component. If one of these attributes is present, the contents of the button will be replaced with a loading indicator while a network request is out.

CleanShot 2024-09-29 at 22 04 52


DonCamillo11:
Not working for me. Updated to 1.0.4 but the loading indicator won’t show up and the text is still visible. Tested also the way from your newsletter https://gist.github.com/calebporzio/ff706c4c646b4f3afa76fb2549737ba8 but it doesn't work either.

christianheidorn:
I'm running into an issue which I'm not sure whether it's intentional or not. These buttons are nested inside rows of a table. I'm using wire:click="..." and the loading indicator works. However, it's triggering all of the other buttons as well.

Is there any way to prevent this behavior?

Monosnap Image 2024-10-01 17-15-10

helderneves91:
> I'm running into an issue which I'm not sure whether it's intentional or not. These buttons are nested inside rows of a table. I'm using `wire:click="..."` and the loading indicator works. However, it's triggering all of the other buttons as well.
> 
> Is there any way to prevent this behavior?
> 
> Monosnap Image 2024-10-01 17-15-10

I'm experiencing the same.

joshhanley:
@christianheidorn @helderneves91 you should add `wire:target` to the buttons to target your specific rows.

christianheidorn:
@joshhanley Thanks, lifesaver! It works now. I just started digging deeper into Livewire so this was confusing the heck out of me.

joshhanley:
@christianheidorn no worries! Glad it worked for you 🙂

PR #250, opened 2024-10-02
Fix button and input group styling

calebporzio:
We did a deep dive and made sure all button and input variants and combinations look good inside groups:

CleanShot 2024-10-02 at 11 35 57@2x

### DIFF
--- a/stubs/resources/views/flux/button/group.blade.php
+++ b/stubs/resources/views/flux/button/group.blade.php
@@ -2,9 +2,9 @@
 $classes = Flux::classes()
     ->add('flex group/button')
     ->add([ // Make the first, middle, and last buttons have proper border radiuses...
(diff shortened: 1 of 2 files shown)


benjibee:
@calebporzio I feel like you _may_ have missed one here, if I understand the documentation correctly.

```blade
<flux:button.group>

    <flux:button size="sm" href="#">Edit</flux:button>

    <flux:dropdown>

        <flux:button size="sm" icon="chevron-down"></flux:button>

        <flux:menu class="min-w-32">
            <flux:menu.item variant="danger">Delete</flux:menu.item>
        </flux:menu>

    </flux:dropdown>

</flux:button.group>
 ```

Results in: 

image


A fix is to add `class="rounded-l-none border-l-0"` to the button inside of the dropdown but it feels like this should "just work" in line with the updates in this PR.

PR #300, opened 2024-10-04
Flux tag compiler class readability cleanup

kyledoesdev:
I was interested in how the Flux Tag Compiler worked and was reading through the class and thought the patterns being in line in each method was a bit difficult to read because of how long the regex was.

I pulled the regex tag compilation patterns out into a helper class called regex in a new directory namespaced
Flux\Helpers. Each pattern is accessible through a static public method that returns the string.

I also cleaned up some unused class & function imports.

PS. The regex pattern is really interesting, I read through it a while and still am unsure how it really works. kudos!


joshhanley:
@kyledoesdev thanks for the PR!

We like to keep PRs specific to the change that is being proposed. As such, can you please revert the unused class and function imports and move them to a new PR? Same with the README fix up.

The reason request this, is because we could merge the import clean ups or the README fix straight away without them having to wait for whether we think this PR might be suitable or not.

Regarding this PR, I personally think how it was is fine. But I can understand wanting to make the methods easier to read.

I think extracting to methods inside the `FluxTagComplier` would be better than extracting to a helper class. This is because the regexs aren't used anywhere else, so it's nice being able to just open the one file (FluxTagCompiler) and see how it works.

Could you make all the above changes and let us know when it is ready?

Thanks!

calebporzio:
Yeah, like @joshhanley  said we're generally not interested in PRs like this as we don't know what other changes they introduce. Additionally the current style mirrors laravel's style exactly and I don't want to diverge from that. Thanks anyways.

kyledoesdev:
Thanks for the feedback! I was unaware that the currently implemented style mimics Laravel's style. On that note, I also agree its best to disregard this one then. Thanks for reviewing guys, cheers!

PR #302, opened 2024-10-04
Add --force option to PublishCommand

ThomasEnssner:
Introduced a --force option to the PublishCommand to override existing files and directories.

This enhancement allows users to forcefully publish components even if they already exist.

Updated the logic to check this new option before skipping existing paths.


joshhanley:
@ThomasEnssner thanks for the PR! @calebporzio just tested this and it works great.

joshhanley (approved):
@ThomasEnssner thanks for the PR! @calebporzio just tested this and it works great.

calebporzio:
Awesome, thanks @joshhanley

Issue #51, opened 2024-09-24
[PRO] Dropdown + Profile weird focus state depending on where user clicks


LasseRafn:
If you click the name it appears that the dropdown menu gets focused, but if I click away from the name it does not. Keyboard navigation works for both cases, only difference is that the outline appears when clicking name, but not when clicking blank or chevron.

Browser: Safari 17.6, macOS 14.6.1

https://github.com/user-attachments/assets/98f691e0-4bf1-4698-97bb-631d97da21dc

joshhanley:
@LasseRafn thanks for reporting! For some reason your video isn't working. Can you re-upload it? Thanks!

LasseRafn:
@joshhanley weird! try this, https://share.cleanshot.com/Fk1wHZBC

https://github.com/user-attachments/assets/5f7ea7a7-9445-4a33-88b8-ee6cc079643e

joshhanley:
@LasseRafn thanks for reporting! I have submitted a PR #400 with a fix.

calebporzio:
Thanks @joshhanley!

PR #400, opened 2024-10-11
Remove focus styles so safari doesn't focus menu on first click

joshhanley:
At the moment if you click on a button that opens a `<flux:menu>` in Safari, on the first click, it adds focus styles to the popover. Any subsequent clicks are fine and no focus styles appear.

image

As the `<flux:menu>` has a `tabindex="-1"` when rendered, I assume that the menu should not be focusable. Because of that in this PR, I've removed the focus styles from the `<flux:menu>` component which fixes this issue.

image

Fixes https://github.com/livewire/flux/issues/51

### DIFF
--- a/stubs/resources/views/flux/menu/index.blade.php
+++ b/stubs/resources/views/flux/menu/index.blade.php
@@ -4,6 +4,7 @@ $classes = Flux::classes()
     ->add('rounded-lg shadow-sm')
     ->add('border border-zinc-200 dark:border-zinc-600')
     ->add('bg-white dark:bg-zinc-700')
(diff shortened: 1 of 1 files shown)


ITrunsDE:
These focus applies also to the toast message. 
image

calebporzio:
Normally I avoid outline-none on things because it's usually a workaround for a deeper problem, but in this case browser's (other than Firefox) don't have a way to programmatically focus an element without showing the indicator.

I also don't think this indicator provides any benefit.

However, the toast close button probably should show the focus indicator as it's a more clickable element.

Thanks Josh

Issue #316, opened 2024-10-04
Table not published after flux:publish --all


NickSB:
Hi there,

table.blade.php is not published after command `php artisan flux:publish --all`.

Solution:
Rename and move from `stubs/resources/views/flux/table.blade.php` to `stubs/resources/views/flux/table/index.blade.php`

Greetings,
Nick

helderneves91:
I think the stubs doesn't overwrite existing ones.
Check https://github.com/livewire/flux/pull/302

joshhanley:
@NickSB thanks for reporting! I have submitted a PR #302 to fix this.

PR #401, opened 2024-10-12
Fix publish command to include table

joshhanley:
Currently when `flux:publish` is run, the table folder is being published, but the table blade file isn't.

It is because in Flux Pro there is both a `table` folder and a `table.blade.php` file and the publish command doesn't support matching both.

This PR fixes the publish command to check for a directory and also for a file.

Fixes https://github.com/livewire/flux/issues/316


calebporzio:
Thanks Josh!

PR #435, opened 2024-10-16
add blade-ui-kit support

danielrona:
Added blade-ui-kit support with buk::icon-name

I was wondering if there was any other interest and found https://github.com/livewire/flux/discussions/294 and https://github.com/livewire/flux/discussions/232 since I myself wanted it too I thought I might as well offer this up there too.

Not sure why all icon files had their own classes as they all where 1:1 identical (wich made the find and replace easy though)

This should not break any existing icons.

If you have the blade-ui-kit installed (and in this example the phoshpor-icon-package) you can use the following code.

Trailing on Button
```
<flux:button icon-trailing="buk::phosphor-gift">
    {{ __('Sample') }}
</flux:button>
```
image

Leading on Button
```
<flux:button icon="buk::phosphor-gift">
    {{ __('Sample') }}
</flux:button>
```
image

Simple Icon
```
<flux:icon icon="buk::phosphor-gift-duotone" />
```
image

> [!NOTE]
> Variants will not work on buk:: icons as they are customized svg in the actual file, but blade-hero-icons usually come with unique named variants.


calebporzio:
Thank you for the submission, however I want to keep scope small initially while we clarify and stabilize the product.

PR #621, opened 2024-11-07
Update FluxTagCompiler.php to fix #620

### LINKED ISSUE #620: Update 1.0.21 : Method Illuminate\View\ComponentAttributeBag::all does not exist.
Hi Fluxers!

After the fix update 1.0.21 we have total dashboard error 1.0.21 Method Illuminate\View\ComponentAttributeBag::all does not exist.

vendor / livewire / flux / stubs / resources/ views / flux/ icon/index.blade.php: 12

It will break code in booting. I think this is urgent.

Image

rzv-me:
this update replaces the use of `$attributes->all()` with `$attributes->getAttributes()` to fix the breaking change from #620

### DIFF
--- a/src/FluxTagCompiler.php
+++ b/src/FluxTagCompiler.php
@@ -18,7 +18,7 @@ class FluxTagCompiler extends ComponentTagCompiler
     'view' => md5('flux') . '::' . {$component},
     'data' => \$__env->getCurrentComponentData(),
 ])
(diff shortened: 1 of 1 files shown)


rzv-me:
This is kindof urgent, is there anyone else aside from @calebporzio, that can merge this? cc: @bensherred

PR #943, opened 2024-12-30
Add `success` variant to button component

aptcreativedesign:
This PR adds the variant "success" to buttons.

```blade
<flux:button type="submit" variant="success">Success</flux:button>
``` 

image

I understand that the "primary" variant is what is usually a success/confirmation.
However there have been times when I felt that the visual cue of a green button would be helpful to the user


joshhanley (approved):

gustavosobrinho01:
This PR is very important, I will definitely use it.

calebporzio:
Thanks for the PR! I'm pretty protective of the button variants and keeping them slim. This would create a precedent I'm not sure I'm willing to support for other components and also might have styling implications for other theming. Going to close for now. Sry, thanks.

Issue #887, opened 2024-12-18
Button icon is bigger using "icon-variant"


magarrent:
### Flux version

v1.1.0

### Livewire version

v3.5.12

### What is the problem?

Hey!!!

I've just discovered a small UI bug in the button component.

When using **icons** with **icon-variant**, let's say **outline**, the icon size is set to **size-6**, instead of the default (4), and looks so big:

Using `icon-variant="outline"`
<img width="268" alt="Image" src="https://github.com/user-attachments/assets/aa9b01df-cb55-4de4-b8de-8265b4acd428" />

**Without** it
<img width="203" alt="Image" src="https://github.com/user-attachments/assets/66a52b17-d3ec-40f8-98ab-e8cc48efd47b" />

Inspecting the console I see that the size changed:

**shrink-0 [:where(&)]:size-6**
_shrink-0 [:where(&)]:size-4_

Hope it's a tiny one! Thanks for all! 🚀

### Code snippets

```
<flux:button type="submit"
             variant="primary"
             icon="bolt"
             icon-variant="outline">
    Start generating ✨
</flux:button>
```

```
<flux:button type="submit"
             variant="primary"
             icon="bolt">
    Start generating ✨
</flux:button>
```

### How do you expect it to work?

Fixing the icon size

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

joshhanley:
@magarrent thanks for reporting! I've submitted a PR with a fix.

PR #987, opened 2025-01-14
Fix outline icon sizes on buttons to ensure they match default icon sizes

joshhanley:
# The scenario

If you use buttons with `icon-variant="outline"` the icons are bigger than the default solid sizes.

image

```blade
<flux:button
    variant="outline"
    icon="bolt"
    icon-trailing="academic-cap"
    icon-variant="outline">
    Outline
</flux:button>
<flux:button
    variant="outline"
    icon="bolt"
    icon-trailing="academic-cap">
    Solid
</flux:button>
```

And the same issue happens with square buttons.

image

# The problem

The issue is that outline icons have a 24x24 view box and there isn't a smaller outline variant. Where as for buttons the default icon variant is micro which has a view box of 16x16.

# The solution

This PR resizes the outline icon variant to be the same size as the default button icon sizes, using `size-4` by default.

image

If the `square` button attribute is used, then the size is set to `size-5`.
image

It's not an ideal solution as the size of the icons is now not encapsulated in the icon components.

An alternative solution to this would be to add sizes as an option on the icons themselves, but that would require a lot more work, as we now need to determine what happens with the solid/mini/micro solid variants.

Ideally in the future we would just have a outline and solid variants and a separate size attribute.

Fixes livewire/flux#887

### DIFF
--- a/stubs/resources/views/flux/button/index.blade.php
+++ b/stubs/resources/views/flux/button/index.blade.php
@@ -23,6 +23,11 @@ $iconVariant ??= ($size === 'xs')
     ? ($square ? 'micro' : 'micro')
     : ($square ? 'mini' : 'micro');
 
+// When using the outline icon variant, we need to size it down to match the default icon sizes...
+$iconClasses = Flux::classes()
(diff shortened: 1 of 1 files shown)


calebporzio:
Can you PR this solution to any component that uses "icon"? plz?

PR #1043, opened 2025-01-28
Add nonce to dark mode script tag

### LINKED ISSUE #1034: Injected script to manage dark mode does not comply with CSP
### Flux version

v1.1.3

### Livewire version

v3.5.17

### What is the problem?

I have a fairly strict CSP and I have disabled dark mode.

But the browser still complains about the script injected by Flux
```
Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' 'nonce-pUT16AksAMo5zjRjbkiC8r9adYK9uLENJMpcYSZD' https://www.googletagmanager.com 'unsafe-eval'". Either the 'unsafe-inline' keyword, a hash ('sha256-10Dh1l13zor4I3gkPqw4Y1m3FpiMnB1VtjcVaCLTYFg='), or a nonce ('nonce-...') is required to enable inline execution.
```

I tried adding an `unsafe-inline` keyword for `script-src` (not that I would want to anyway) but Firefox says it's ignored anyway because a "nonce value is present in the source list", which is true as we do use nonce.

### Code snippets

`taiwind.config.js`
```js
import defaultTheme from 'tailwindcss/defaultTheme'

/** @type {import('tailwindcss').Config} */
export default {
  darkMode: null,
  ...
}
```

`app.blade.php`
```blade
<!DOCTYPE html>
<html lang="{{ str_replace('_', '-', app()->getLocale()) }}">
<head>
    ...
    @livewireStyles(['nonce' => csp_nonce()])
    @vite('resources/css/app.css')
    @fluxStyles()
</head>
<body class="min-h-screen">
    ...
@fluxScripts()
@vite('resources/js/app.js')
@livewireScripts(['nonce' => csp_nonce()])
</body>
</html>
```

Since I disabled dark mode I would expect Flux not to inject the script in the first place.

However, since the script will be necessary when dark mode is selectable, I would expect a way to add a `nonce` value, maybe similarly to what Tailwind does by adding a configuration to the blade directive

```balde
@fluxScripts(['nonce' => csp_nonce()])
```

joshhanley:
# The scenario

If you have a strict CSP setup, an error is being thrown because of Flux's dark mode detection script that gets injected with the styles.

```
Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' 'nonce-pUT16AksAMo5zjRjbkiC8r9adYK9uLENJMpcYSZD' https://www.googletagmanager.com 'unsafe-eval'". Either the 'unsafe-inline' keyword, a hash ('sha256-10Dh1l13zor4I3gkPqw4Y1m3FpiMnB1VtjcVaCLTYFg='), or a nonce ('nonce-...') is required to enable inline execution.
```

# The problem

The issue is that Flux's styles and scripts both support a nonce but the new inline dark mode script does not.

# The solution

The solution is to add the nonce that can be passed into `@fluxStyles(['nonce' => 'abc'])` to the inline script.

Fixes #1034

### DIFF
--- a/src/AssetManager.php
+++ b/src/AssetManager.php
@@ -107,7 +107,7 @@ class AssetManager
 
         return <<<HTML
 <link rel="stylesheet" href="/flux/flux.css?id=$versionHash"$nonce>
-<script>
+<script$nonce>
     let appearance = window.localStorage.getItem('flux.appearance') || 'system'
 
     if (appearance === 'system') {


troccoli (commented):
Are you sure you don't need a space?
```html
<script $nonce>
```

joshhanley:
@troccoli thanks for checking! Nah the space is built into the `$nonce` variable.

```php
$nonce = isset($options) && isset($options['nonce']) ? ' nonce="' . $options['nonce'] . '"' : '';`
```

troccoli:
Thank you @joshhanley for fixing it quickly

Issue #1191, opened 2025-02-25
Inconsistent font size on various components


0hra:
### Flux version

v2.0.2

### Livewire version

v3.5.19

### Tailwind version

v4.0.7

### Browser and Operating System

Firefox/Safari on Macos

### What is the problem?

Font-size isn't consistent between components. Tested on fresh Livewire Starter Kit.

https://github.com/user-attachments/assets/27747cc1-d576-41e3-9a37-f03c65fbbe04

### Code snippets

```blade
<div class="flex flex-col gap-2">
        <flux:input label="Input" value="Feb 5, 2025" />
        <flux:date-picker label="Date picker" value="2025-02-05"/>
        <flux:select label="Select">
            <flux:select.option selected>Feb 5, 2025</flux:select.option>
        </flux:select>
</div>
```

### How do you expect it to work?

Consistent font sizes.

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

joshhanley:
@0hra thanks for reporting! I've submitted a PR with a fix.

PR #1197, opened 2025-02-26
Fix select sizes to match other inputs

joshhanley:
# The scenario

Currently if you have inputs, date pickers and selects together in the same form, they are displaying different text sizes

**Mobile before**
image

**Desktop before**
image

```blade
<?php

use Livewire\Volt\Component;

new class extends Component {
    //
}; ?>

<div class="flex flex-col gap-2">
    <flux:input label="Input" value="5 Feb 2025" />
    <flux:date-picker label="Date picker" value="2025-02-05"/>
    <flux:date-picker value="2025-02-05">
        <x-slot name="trigger">
            <flux:date-picker.input />
        </x-slot>
    </flux:date-picker>
    <flux:select label="Select">
        <flux:select.option selected>5 Feb 2025</flux:select.option>
    </flux:select>
    <flux:select variant="listbox" label="Listbox">
        <flux:select.option selected>5 Feb 2025</flux:select.option>
    </flux:select>
    <flux:select variant="combobox" label="Combobox">
        <x-slot name="input">
            <flux:select.input value="5 Feb 2025" />
        </x-slot>
        <flux:select.option selected>5 Feb 2025</flux:select.option>
    </flux:select>
</div>
```

# The problem

The issue is that selects and date pickers didn't have the correct text sizes applied like inputs do.

The sizes we have on inputs for the `default` size are:

```php
default => 'text-base sm:text-sm ...'
```

These were applied in a previous PR where we decided to make mobile 16px (text-base) by default and 14px (text-sm) for desktops.

# The solution

This PR fixes selects and the date picker to have the correct sizes for the default size.

There is also a corresponding Flux Pro PR livewire/flux-pro#199

**Mobile after fix**
image

**Desktop after fix**
image

Fixes livewire/flux#1191

### DIFF
--- a/stubs/resources/views/flux/select/variants/default.blade.php
+++ b/stubs/resources/views/flux/select/variants/default.blade.php
@@ -12,7 +12,7 @@ $classes = Flux::classes()
     ->add('appearance-none') // Strip the browser's default <select> styles...
     ->add('w-full pl-3 pr-10 block')
     ->add(match ($size) {
-        default => 'h-10 py-2 text-sm leading-none rounded-lg',
+        default => 'h-10 py-2 text-base sm:text-sm leading-none rounded-lg',
         'sm' => 'h-8 py-1.5 text-sm leading-none rounded-md',
         'xs' => 'h-6 text-xs leading-none rounded-md',
     })


calebporzio:
great - thank you Josh

PR #1202, opened 2025-02-27
Fix deprecation warning when button has missing wire:click

Fludem:
I just changed the conditional that sets the $isJsMethod value to use an empty string in cases where there isn't a wire:click attribute passed to the button component.

It may not be the best solution; I haven't really looked at the rest of the codebase. A user in the Livewire Discord posted that he'd received these notices, and I'm assuming it's from Flux.

Discord Details:

[Original Message ](https://discord.com/channels/698229985755791471/698229986233942028/1344385903233335390)

image


Fludem:
I haven't raised an issue as I can't 100% be certain these warnings are coming from flux's button component.

calebporzio:
Thanks

Issue #1209, opened 2025-02-27
Missing options in <flux:select> when “variant” is not specified


MOA-TI:
### Flux version

v2.0.3

### Livewire version

v3.6.0

### Tailwind version

v4.0.8

### Browser and Operating System

Chrome and macOS

### What is the problem?

When using <flux:select> without specifying the "variant" attribute, the options inside the component are not displayed. This issue occurs even though the options are correctly provided in the Blade template. Adding a "variant" explicitly resolves the problem, suggesting that the default behavior may not be working as expected.

### Code snippets

```php
    <flux:select wire:model.live="concepto_id"
                 label="Concepto"
                 required
                 descriptionTrailing="Sólo se muestran los conceptos activos para el trabajador."
                 placeholder="Buscar...">
        @foreach($conceptos as $k => $v)
            <flux:select.option :value="$k">{{ $v }}</flux:select.option>
        @endforeach
    </flux:select>
```

This code produces this output
<img width="895" alt="Image" src="https://github.com/user-attachments/assets/e2325613-6681-4761-8d64-b68d580ca605" />

If you explicitly define variant="default" in the flux:select tag (which is the default value), then the generated code is:
<img width="561" alt="Image" src="https://github.com/user-attachments/assets/c93e0f92-7882-4c0b-b8d8-973dec6eebd6" />

### How do you expect it to work?

This worked perfectly in version 1.

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

jeffchown:
@joshhanley I have had this happen to me when the select is in a modal.  I will double-check and provide a Volt component tomorrow if I can reproduce the issue.

joshhanley:
@jeffchown ah thanks! You are correct it seems when in a modal that has a variant prop defined, then they don't show up. Will reopen.

Volt component that demonstrates the issue.

```blade
<?php

use Livewire\Volt\Component;

new class extends Component {
    public $items = [
        'item1' => 'Item 1',
        'item2' => 'Item 2',
        'item3' => 'Item 3',
    ];
}; ?>

<div>
    <flux:modal.trigger name="edit-profile">
        <flux:button>Edit profile</flux:button>
    </flux:modal.trigger>

    <flux:modal variant="flyout" name="edit-profile" class="md:w-96">
        <div>
[...]

PR #1213, opened 2025-02-28
Fix select option variant not found

joshhanley:
# The scenario

If you use a default select inside a component that has a variant defined, like a modal, then the select options aren't showing up.

image

```blade
<?php

use Livewire\Volt\Component;

new class extends Component {
    //
}; ?>

<div>
    <flux:modal.trigger name="edit-profile">
        <flux:button>Edit profile</flux:button>
    </flux:modal.trigger>

    <flux:modal variant="flyout" name="edit-profile" class="md:w-96">
        <flux:select placeholder="Choose industry...">
            <flux:select.option>Accounting</flux:select.option>
            <flux:select.option>Legal services</flux:select.option>
            <flux:select.option>Consulting</flux:select.option>
            <flux:select.option>Other</flux:select.option>
        </flux:select>
    </flux:modal>
</div>
```

# The problem

The issue is that we recently split the select components, so some are free and some are pro, which required the option component to be split into multiple variants, but I didn't add the check back in to make sure that the variant for the option is coming from a select component and not any other components.

```blade
@php
// If the variant is coming from the parent component and is not default, we need to use the custom variant...
$variant = $variant === 'default' ? 'default' : 'custom';
@endphp
```

# The solution

The solution is to update the variant check to include a check to make sure the select variant exists first. Now everything works happily.

```blade
@php
// This prevents variants picked up by `@aware()` from other wrapping components like flux::modal from being used here...
$variant = $variant !== 'default' && Flux::componentExists('select.variants.' . $variant)
    ? 'custom'
    : 'default';
@endphp
```

image

Fixes livewire/flux#1209

### DIFF
--- a/stubs/resources/views/flux/select/option/index.blade.php
+++ b/stubs/resources/views/flux/select/option/index.blade.php
@@ -5,8 +5,10 @@
 ])
 
 @php
-// If the variant is coming from the parent component and is not default, we need to use the custom variant...
-$variant = $variant === 'default' ? 'default' : 'custom';
+// This prevents variants picked up by `@aware()` from other wrapping components like flux::modal from being used here...
+$variant = $variant !== 'default' && Flux::componentExists('select.variants.' . $variant)
+    ? 'custom'
+    : 'default';
 @endphp
 
 <flux:with-field :$attributes>


Fludem:
Will you write my PRs descriptions for me

calebporzio:
legendary descriptions eh @Fludem? thanks Josh!

phikhi:
I ran composer update in my project and the problem still persists but only when you set the modal variant to `flyout`.

with the default modal, the select options are showing as expected. 
But not in the flyout variant.

Is it just me ?

joshhanley:
@phikhi can you run `composer show -D` and screenshot the output please?

phikhi:
Of course, here you go ;)

Capture d’écran 2025-03-10 à 11 40 26

joshhanley:
@phikhi all looks ok! I’ve seen Flux not upgrade before due to Livewire versions. Can you also try and run `php artisan view:clear` and see if that fixes it?

phikhi:
It solved it perfectly... I should've tried that before ! 
Sorry and thanks for your help 😉

joshhanley:
@phikhi no worries! 🙂 I’m just glad it was a simple fix 😁

PR #1217, opened 2025-02-28
Switch Toggle: Replace w-8 with min-w-8 to fix parent's flex issues

dimitri-koenig:
I have a switch toggle placed within a flex and justify-between element, which get's truncated and looks like this:

Bildschirmfoto 2025-02-28 um 09 40 08

The only fix I found so far is to replace the fixed width with a min width, so instead of w-8 we take min-w-8, and voila, it works:

Bildschirmfoto 2025-02-28 um 09 40 16

For clarification, my blade template is this:
```blade
<div class="flex items-start gap-4 justify-between">
    <flux:heading>{{ __('Get email notifications about new subscribers') }}</flux:heading>
    <flux:switch wire:model.live="state.notify_about_new_subscriber" class="cursor-pointer"/>
</div>
```

### DIFF
--- a/stubs/resources/views/flux/switch.blade.php
+++ b/stubs/resources/views/flux/switch.blade.php
@@ -5,7 +5,7 @@
 
 @php
 $classes = Flux::classes()
-    ->add('group h-5 w-8 relative inline-flex items-center outline-offset-2')
+    ->add('group h-5 min-w-8 relative inline-flex items-center outline-offset-2')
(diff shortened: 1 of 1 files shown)


calebporzio:
Ah good call - we should actually still keep `w-8` I think as it really shouldn't exceed that width.

seabasss:
Sorry if I'm just spitting out stuff, but isn't this what shrink-0 is for?

PR #1228, opened 2025-03-02
Brand Component Improvements

Fludem:
# Styling Improvements with some refactoring.

I've just started using Flux, and I needed to modify the Brand component in my app.

I noticed that my logo was rendering too small on my side-nav. It was also at the top of its parent container, while the text was vertically centered, breaking the alignment.

## Fixed Image Alignment 

Example:

image

Fixed by adding inline-flex and align-middle to the img element's parent container.
object-contain was also added to the img element to prevent the brand image from being vertically stretched.

image

## Text is misaligned when no logo is present

If a logo is rendered with a name but no logo, the logo's fixed size parent is still rendered, causing the name to be misaligned. I've made the container red to make it visible:

image

This element now has the hidden class if no logo is provided to prevent it from mis-aligning the brand name.

## Refactoring

The conditional code handling how the component is rendered when the name is/isn't null seemed redundant. The only changes between the two outputs were the logo image size and the name being printed. Thought it would make sense to merge them both together and remove the repeated code.

### DIFF
--- a/stubs/resources/views/flux/brand.blade.php
+++ b/stubs/resources/views/flux/brand.blade.php
@@ -5,35 +5,29 @@
 ])
 
 @php
-$classes = Flux::classes()
-    ->add('h-10 flex items-center mr-4')
-    ;
+    $classes = Flux::classes()->add('h-10 flex items-center mr-4');
 
-$textClasses = Flux::classes()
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks for the contribution however, it's too difficult to tell what has changed here for me to be comfortable merging this. Please separate out your changes and make them as minimal as possible for the outcome you're trying to achieve. Thanks!

Issue #983, opened 2025-01-13
Flux:select selected not working correctly


ITrunsDE:
### Flux version

v1.1.2

### Livewire version

v3.5.18

### What is the problem?

Hi,

I have a flux modal with a small form. That shows three rows with a select in the front and a text field. The select is a normal dropdown with the values "lt", "le", "eq", "ge", "gt". 

The first row with the select is working correctly, but the second row, the select is not working. Even when I try to set to use the same wire:model. 

Is there something that I missed?

When I try this with basic HTML select, the options are selected correctly.

Best Regards
Sebastian

### Code snippets

```blade
<flux:modal name="edit-subscription" class="md:w-128 space-y-6">
    <div>
        <flux:heading size="lg">Edit your subscription for {{ $editSubscription->product->name ?? '' }}</flux:heading>
    </div>
    <flux:fieldset>

        <flux:switch wire:model="subscriptionForm.active" label="Enable notifications"/>

        <flux:legend>Filter Major Version</flux:legend>

        <div class="space-y-6">
            <div class="grid grid-cols-2 gap-4">
                <flux:select size="sm" label="Condition" wire:model="subscriptionForm.major_symbol" :filter="false">
                    <flux:option>unset</flux:option>
                    <flux:option value="lt">&lt;</flux:option>
                    <flux:option value="le">&le;</flux:option>
                    <flux:option value="eq">&equals;</flux:option>
                    <flux:option value="ge">&ge;</flux:option>
                    <flux:option value="gt">&gt;</flux:option>
                </flux:select>
                <flux:input wire:model="subscriptionForm.major" size="sm" label="Version"/>
            </div>
        </div>
    </flux:fieldset>

    <flux:separator/>

    <flux:fieldset>
        <flux:legend>Filter Minor Version</flux:legend>

        <div class="space-y-6">
            <div class="grid grid-cols-2 gap-4">
                <flux:select size="sm" label="Condition" wire.model="subscriptionForm.minor_symbol" :filter="false">
                    <flux:option>unset</flux:option>
                    <flux:option value="lt">&lt;</flux:option>
                    <flux:option value="le">&le;</flux:option>
                    <flux:option value="eq">&equals;</flux:option>
                    <flux:option value="ge">&ge;</flux:option>
                    <flux:option value="gt">&gt;</flux:option>
                </flux:select>
                <flux:input wire:model="subscriptionForm.minor" size="sm" label="Version"/>
            </div>
        </div>
    </flux:fieldset>

[...]

PR #1234, opened 2025-03-03
Add the selected prop to select options

Harry-Harrison:
This adds a selected prop to the select options components so that the user can force an option as selected properly. 

Saw a few people reporting this issue and couldn't find anyone with a fix.

- https://github.com/livewire/flux/discussions/1089
- https://github.com/livewire/flux/issues/983

## Example
```blade
<flux:select variant="listbox" placeholder="Select Something..." wire:change="setSomething($event.target.value)">
    @foreach($somethings as $something)
        <flux:select.option :selected="$selectedSomething === $something->id" :value="$something->id">
            {{ $something->name }}
        </flux:select.option>
    @endforeach
</flux:select>
```

### DIFF
--- a/stubs/resources/views/flux/select/option/variants/default.blade.php
+++ b/stubs/resources/views/flux/select/option/variants/default.blade.php
@@ -1,9 +1,11 @@
 @props([
     'value' => null,
+    'selected' => null,
 ])
 
 <option
     {{ $attributes }}
     @if (isset($value)) value="{{ $value }}" @endif
     @if (isset($value)) wire:key="{{ $value }}" @endif
+    @if ($selected) selected @endif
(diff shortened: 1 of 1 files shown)


joshhanley:
@Harry-Harrison thanks for the PR! The option doesn't need `selected` as a prop as it already works.

I just tested this and it works as expected.

```blade
<flux:select placeholder="Choose industry...">
    <flux:select.option selected>Accounting</flux:select.option>
    <flux:select.option>Legal services</flux:select.option>
    <flux:select.option>Consulting</flux:select.option>
    <flux:select.option>Other</flux:select.option>
</flux:select>
```

image

I'm going to close this, but if it doesn't work for you, please let me know and I can re open it 🙂

PR #1305, opened 2025-03-11
Make Component Search Case-Insensitive When Publishing

sinlyu:
# The scenario

When publishing a component, the search functionality only works correctly when the input is in lowercase.

# The problem

If the user enters uppercase characters in the search input, no results are returned:
image

# The solution

This PR updates the `searchOptions` method in `PublishCommand` and normalizes the `$value` to lowercase when filtering results.

With this change, the search becomes case-insensitive and works correctly regardless of input casing:

image


joshhanley (approved):
@sinlyu thanks for the PR! @calebporzio I've just tested this locally and all looks good.

calebporzio:
Thanks!

Issue #1222, opened 2025-03-01
Date-picker range mode to display validation error


omniith:
### Flux version

v2.03

### Livewire version

v3.6

### Tailwind version

v4.0.7

### Browser and Operating System

Chrome on macOS

### What is the problem?

Flux date-picker range does not display the error on startend. I am using date-picker mode range and need to do validation on the dates, currently using rules in livewire forms. Though currently flux ensures start and end is in $range, I also need to do custom validations that require a display on error if start/end fails to meet validation.

### Code snippets

```blade
protected function rules()
{
    return [
        'range.start' => ['required', new DateNotOverlapping()],
        'range.end' => ['required', new DateNotOverlapping()],
    ];
}
```
DateNotOverlapping
```
public function validate(string $attribute, mixed $value, Closure $fail): void
{
    if($this->user->leaves()
        ->where('start_date', '<=', $value)
        ->where('end_date', '>=', $value)
        ->exists()) {
        $fail('You have taken absence during these days.');
    }
}
```
### How do you expect it to work?

To be able to display errors for range start end, will be great if documentation can give an example on what custom validations should expect. (I'm assuming flux is completely on Laravel+Livewire or Laravel+Livewire+Volt)

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

joshhanley:
Imagine that we have a form and we have a `DatePicker` in `range` mode with presets enabled.

```blade
<flux:date-picker mode="range" label="Range" wire:model.live="range" with-presets />
```

This is mapped to a range public property.

```php
public ?DateRange $range = null;
```

Now, we want to add some validation to ensure that the data is coming from the front end correctly and hasn't been tampered with.

The first instinct would be to add the validate attribute.

```php
#[Validate(['required'])]
public ?DateRange $range = null;
```

We add required first as we want to ensure that it's correctly filled out.

But then how do we ensure that the `DateRange` contains valid start and end dates, that the end date is after start date, and if the user has selected a preset - that it's a valid `DateRangePreset` option.

As Livewire can't validate nested property values using the validate attribute, instead we need to resort to using the rules method and come up with something like this.

```php
public function rules()
{
    return [
        'range.start' => ['required', 'date'],
        'range.end' => ['required', 'date', 'after:range.start'],
[...]

PR #1308, opened 2025-03-12
Update error component to support nested attributes

joshhanley:
While this PR doesn't actually depend on PR #1295, I have branched this PR off that one to make testing easier, so I would recommend merging that PR first before reviewing this one. 

# The scenario

Currently if we have a component, like the date picker, that updates a property which has nested attributes, then the validation errors for those nested attributes aren't shown.

For example, if we have a date range picker and we have rules for start and end to ensure that start is before end, then if a range is submitted the wrong way around and validation fails, no error messages are show for `range.start` or `range.end`.

image

```blade
<?php

use App\Rules\DateRangeRule;
use Flux\DateRange;
use Flux\DateRangePreset;
use Livewire\Attributes\Computed;
use Livewire\Attributes\Validate;
use Livewire\Volt\Component;

new class extends Component {
    public ?DateRange $range = null;

public function rules()
{
    return [
        'range' => ['required'],
        'range.start' => ['required', 'date', 'before:range.end'],
        'range.end' => ['required', 'date', 'after:range.start'],
        'range.preset' => ['nullable', 'in:' . collect(DateRangePreset::cases())->pluck('value')->implode(',')],
    ];
}

    public function submit()
    {
        $this->validate();

        Flux::toast('Date range is valid!', variant: 'success');
    }
};
?>

<div>
    <form wire:submit="submit">
        <flux:date-picker mode="range" label="Range" wire:model.live="range" with-presets />
        <flux:button variant="primary" type="submit">Submit</flux:button>
    </form>
</div>
```

# The problem

The issue is that the errors component is only showing exact errors for `range` and not any of range's attributes, like `range.start`.

# The solution

This PR adds support to the `<flux:error>` component for showing errors for the property and any nested attributes. So the below code, will display errors for the range property itself.

```blade
<flux:error name="range" />
```
Screenshot 2025-03-12 at 1 12 37 pm

And if there are no errors on the base property, it will display any errors on nested attributes for the property.

Screenshot 2025-03-12 at 1 13 00 pm

It will only display the first nested attribute error it comes across. So if there are errors on both `range.start` and `range.end`, then it will only show the `range.start` error message.

In case someone doesn't want the errors component to display nested attribute errors, they can set `:nested="false"`.

```blade
<flux:error name="range" :nested="false" />
```

This will then only show errors for `range` and not `range.start`, etc.

Fixes livewire/flux#1222

And if you noticed that the red border is missing from the date picker when the error is a nested attribute error, that is because the date picker itself isn't using the nested logic, so I've also submitted a PR to fix that https://github.com/livewire/flux-pro/pull/228

### DIFF
--- a/stubs/resources/views/flux/error.blade.php
+++ b/stubs/resources/views/flux/error.blade.php
@@ -7,7 +7,7 @@
 @php
 $message ??= $name ? $errors->first($name) : null;
 
-if ((is_null($message) || $message === '') && $nested === true) {
+if ((is_null($message) || $message === '') && filter_var($nested, FILTER_VALIDATE_BOOLEAN, FILTER_NULL_ON_FAILURE) !== false) {
     $message = $errors->first($name . '.*');
 }
 
--- a/stubs/resources/views/flux/error.blade.php
+++ b/stubs/resources/views/flux/error.blade.php
@@ -1,10 +1,15 @@
 @props([
     'name' => null,
     'message' => null,
+    'nested' => true,
 ])
 
 @php
 $message ??= $name ? $errors->first($name) : null;
+
+if ((is_null($message) || $message === '') && $nested === true) {
+    $message = $errors->first($name . '.*');
+}
 
 $classes = Flux::classes('mt-3 text-sm font-medium text-red-500 dark:text-red-400')
     ->add($message ? '' : 'hidden');


calebporzio:
that filter_var() bit is slick - thanks Josh

PR #1340, opened 2025-03-17
Add RTL support

joshhanley:
# The scenario

Currently Flux doesn't work for locales that are RTL

# The proposal

This PR adds support for RTL to Flux and has a corresponding Flux Pro PR livewire/flux-pro#230.

I had to make some decisions while adding support for this, which are documented below.

- Breadcrumbs
	- Made separator null and use chevron-right and chevron-left with hide/show based on RTL as default
	- For slash there is no backslash icon, so using scale to flip instead
	- Made separator slottable so people can configure it themselves
- Submenu
	- Default is trailing icon with chevron-right, so added a chevron-left and hide/show based on RTL
	- This is already slottable so people can configure
	- Have updated the floating-ui logic to show submenu right by default and left for RTL.
	- Added an RTL javascript utility function for detecting RTL for floating-ui.
- Label
	- I've made it an inline-flex so it can be reversed automatically by RTL so the badge is on the left
- Loading icon
	- It's still spinning clockwise, ChatGPT thinks that's ok
- Tooltip
	- ~~I've left the tooltip for now. But default it's position top anyway.~~
	- ~~Someone RTL may want right/left to be reversed, or changed to start/end, but going to leave it for now and wait for feedback.~~
	- Floating UI doesn't support start/ end natively, so I've added support for `position="start"` and `position="end"` and mapped them to left/right accordingly depending on RTL.
- Charts
        - Have been left for now.
- Date picker/ Calendar
        - Have been updated to support it.

Fixes livewire/flux#196


3adeling:
Thanks @joshhanley for this PR. I tested all the components and below are my findings:

1. Breadcrumb ✅
2. Submenu: the submenu should be opened to the left side not right

image

3. Loading icon: being clockwise is fine nothing related to RTL.
4. Tooltip: In my opinion, using "start" and "end" is better than "left" and "right" because, for multilingual sites, adjusting left and right dynamically based on the site language can be challenging. Instead, "start" will position the tooltip to the left in LTR languages and to the right in RTL languages, while "end" will do the opposite.
5. Charts: charts being LTR is ok for RTL languages.
6. Calendar: couldn't test the calendar component due to conflict when activating flux pro.

joshhanley:
Thanks for testing @3adeling!

Awesome! Yeah re the submenu, I've got it opening left, I just didn't include the compiled assets in the PR, so that should be good when it gets released 🙂

Re the tooltip - ok, I'll see what Caleb thinks about fixing that. At the moment Floating UI only supports left/right but I'm sure we could do our own start/end translation in the middle and map it to left/right.

Calendar - yep, I couldn't find an easy way to share Flux Pro RTL branch. I'll see if I can work something out there. This is what it looks like.

image

joshhanley:
I've added changes to the tooltip to support `position="start"` and `position="end"`. It's not officially supported by Float UI though, so not sure what Caleb will think of that. I've also committed the built assets for testing.

Issue #1286, opened 2025-03-10
Loadable attribute for inputs


benjibee:
The documentation has quite an elegant example of an inset loader on an input:

```blade
<flux:input icon-trailing="loading" placeholder="Search transactions" />
```
I might be missing something but how exactly would this be used with Livewire? I can use `<flux:field>` here and recreate the icon element from the input component as such:

```blade
<flux:field>
    <flux:label>Search</flux:label>

    <div class="relative">
        <flux:input wire:model.live="search" placeholder="Search transactions" />

        <div wire:loading class="pointer-events-none absolute top-2 bottom-0 flex items-center justify-center text-xs text-zinc-400/75 pr-3 right-0">
            <flux:icon icon="loading" variant="mini" />
        </div>
    </div>

    <flux:error name="search" />
</flux:field>
```
But this doesn't seem like a great solution. I could imagine something along the lines of `copyable`, `viewable`, etc.

```blade
<flux:input wire:model.live="search" placeholder="Search transactions" loadable />

{{-- With target specified --}}
<flux:input wire:model.live="search" placeholder="Search transactions" loadable="term" />
```

_Originally posted by @benjibee in https://github.com/livewire/flux/discussions/332_

benjibee:
> Good point, here's an API I could envision. What are everyone's thoughts?
> 
> <flux:input wire:model="search" /> -> no loading indicator ever
> <flux:input wire:model.live="search" /> -> loading indicator when request is out
> <flux:input wire:model.live="search" :loading="false" /> -> no loading indicator ever
> <flux:input loading /> -> always show loading indicator
> <flux:input loading="foo" /> -> show loading indicator when a request is out for the foo property (wire:loading + wire:target type deal)

This looks like a good solution to me, and it would solve my use case 👍

joshhanley:
Yep I agree too! Will see if I can get it running 😁

joshhanley:
I've submitted a PR with this implementation

PR #1357, opened 2025-03-20
Add loading support to inputs

joshhanley:
# The scenario

Currently if an input is used for search or something like that, there is no easy way to get the trailing icon to display a spinner when a Livewire request is happening.

```blade
<flux:input wire:model.live="search" icon-trailing="loading" placeholder="Search transactions" />
```

Instead you have to use slots.

```blade
<flux:input wire:model.live="search" placeholder="Search transactions">
    <x-slot:trailing-icon>
        <flux:icon.loading wire:loading wire:target="search" />
    </x-slot>
</flux:input>
```

# The proposal

This PR adds support for the following API mapped out by @calebporzio here https://github.com/livewire/flux/issues/1286#issuecomment-2719092938.

```blade
<flux:input wire:model="search" />{{-- no loading indicator ever --}}
<flux:input wire:model.live="search" />{{-- loading indicator when request is out --}}
<flux:input wire:model.live="search" :loading="false" />{{-- no loading indicator ever --}}
<flux:input loading />{{-- always show loading indicator --}}
<flux:input loading="foo" />{{-- show loading indicator when a request is out for the foo property (wire:loading + wire:target type deal) --}}
```

# Trailing icon/ control issue

**UPDATE:**

I've added support for multiple trailing icons, so now the loading icon just appears to the left of all the other icons. I am calculating the required amount of padding based on how many of the trailing icons/controls are present. I'm also using `wire:loading.class` to increase the padding while loading is happening.

**Without loading**
image

**With loading**
image

The below is no longer relevant.

---

The only issue with the implementation is what should happen if there are currently any trailing icons or extra controls like copyable, clearable, etc.?

image

image

Volt component for testing
```blade
<?php

use Livewire\Volt\Component;

new class extends Component {
    public $search;
    public $foo;

    public function updated()
    {
        sleep(2);
    }

    public function submit()
    {
        sleep(2);
    }
}; ?>

<div>
    <form wire:submit="submit">
        <flux:input as="button">Should there be a loading indicator for this button?</flux:input>
        <flux:input label="No loading indicator ever" wire:model="search" />
        <flux:input label="Loading indicator when request is out" wire:model.live="search" />
        <flux:input label="No loading indicator ever" wire:model.live="search" :loading="false" />
        <flux:input label="Always show loading indicator" loading />
        <flux:input label="Show loading indicator when a request is out for the foo property (wire:loading + wire:target type deal)" loading="foo" />
        <flux:input wire:model="foo" />


        <flux:input label="Keyboard Shortcut" loading kbd="⌘K" />
        <flux:input label="Icon Trailing" loading icon-trailing="credit-card" />
        <flux:input label="Expandable" loading expandable />
        <flux:input label="Clearable" loading clearable />
        <flux:input label="Copyable" loading copyable />
        <flux:input label="Viewable" loading viewable />

        <flux:button type="submit">Submit</flux:button>

        <flux:icon name="loading" wire:loading wire:target="search" class="mt-4" />
    </form>
</div>

```

Fixes livewire/flux#1286

### DIFF
--- a/stubs/resources/views/flux/input/index.blade.php
+++ b/stubs/resources/views/flux/input/index.blade.php
@@ -13,6 +13,7 @@
     'copyable' => null,
     'viewable' => null,
     'invalid' => null,
+    'loading' => null,
     'type' => 'text',
     'mask' => null,
     'size' => null,
@@ -22,6 +23,28 @@
 ])
 
 @php
+
+// There are a few loading scenarios that this covers:
+// If `:loading="false"` then never show loading.
+// If `:loading="true"` then always show loading.
+// If `:loading="foo"` then show loading when `foo` request is happening.
+// If `wire:model` then never show loading.
+// If `wire:model.live` then show loading when the `wire:model` value request is happening.
+$wireModel = $attributes->wire('model');
+$wireTarget = null;
+
+if ($loading !== false) {
+    if ($loading === true) {
+        $loading = true;
+    } elseif ($wireModel?->directive) {
+        $loading = $wireModel->hasModifier('live');
+        $wireTarget = $loading ? $wireModel->value() : null;
+    } else {
+        $wireTarget = $loading;
+        $loading = true;
+    }
+}
+
 $invalid ??= ($name && $errors->has($name));
 
 $iconLeading ??= $icon;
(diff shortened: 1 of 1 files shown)


stahlit:
Nice addition.

"The only issue with the implementation is what should happen if there are currently any trailing icons or extra controls like copyable, clearable, etc.?"

Hide the trailing icon/extra control during loading I think would be a nice solution? 

My first thought atleast.

joshhanley:
@stahlit thanks! I ended up adding support for multiple trailing icons, so they don't overlap, which includes the loading icon.

calebporzio:
This is incredible. thanks Josh

benjibee:
Thank you both! 👏

PR #1384, opened 2025-03-24
Add min and max support to inputs

sinlyu:
# The scenario

Currently if an input is from type number it does not support the min and max attributes.

```html
<flux:input type="number" min="1" max="10" />
```

# The proposal

This PR adds support for the min and max attributes for the input element.


joshhanley:
@sinlyu thanks for the PR!

I've just tested your code snippet on the current version of Flux and the min and max attributes are already being forwarded to the input element as expected, so it's all working for me (see screenshot below). So I'm going to close this. If you're experiencing an issue with this, please open an issue and we'll investigate it further. Thanks for being willing to contribute!

image

Issue #1385, opened 2025-03-24
Laravel 10 - Method Illuminate\View\ComponentAttributeBag::all does not exist


scottzirkel:
### Flux version

v2.1.1

### Livewire version

v3.5.19

### Tailwind version

v4.0.9

### Browser and Operating System

Safari on macOS

### What is the problem?

Laravel 10 does not have an `all` method on the `ComponentAttributeBag`

This is a similar issue to #620 

![Image](https://github.com/user-attachments/assets/e391f8d8-c0d1-4f05-924a-bdafc0bf1237)

### Code snippets

In my case, it was triggered by the profile component:

```
<flux:profile
    class="cursor-pointer"
    initials="SZ"
/>
```

### How do you expect it to work?

I believe changing `FluxManager` at line 127 from `$attributes->all()` to `$attributes->getAttributes()` should solve the issue.

I will have a PR incoming.

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

bridgeyuwa:
But `$attributes->getAttributes()`does not include modifications ( `merge()` `except()` `only()` ) .

PR #1386, opened 2025-03-24
Updates FluxManager to be compatible with Laravel 10

scottzirkel:
Updated FluxManager line 127 to use `->getAttributes()` on the `ComponentAttributeBag` rather than `->all()` to be compatible with Laravel 10.

Fixes livewire/flux#1385
Similar to livewire/flux#621

### DIFF
--- a/src/FluxManager.php
+++ b/src/FluxManager.php
@@ -124,7 +124,7 @@ class FluxManager
     {
         $newAttributes = new \Illuminate\View\ComponentAttributeBag($default);
 
-        foreach ($attributes->all() as $key => $value) {
+        foreach ($attributes->getAttributes() as $key => $value) {
             if (str_starts_with($key, $prefix)) {
                 $newAttributes[substr($key, strlen($prefix))] = $value;
             }


joshhanley (approved):
@scottzirkel thanks for the PR! Please don't delete our PR template and instead fill it in. It's there for a reason 🙂

# The scenario

Currently if you try to use the profile component on Laravel 10, the below error will be thrown.

image

# The problem

The issue is that Laravel 10's `ComponentAttributeBag` doesn't have an `->all()` method.

A similar issue was previously fixed in PR #621.

This new issue was introduced with the addition of the `Flux::attributesAfter()` helper method, which makes use of `->all()`.

# The solution

The solution is to change the `Flux::attributesAfter()` method to use `->getAttributes()` the same as what was changed in PR #621.

scottzirkel:
Ah, sorry @joshhanley I didn't even realize I had done that.

calebporzio:
Thanks for the report @scottzirkel and thanks for demonstrating an amazing writeup @joshhanley

Issue #1375, opened 2025-03-23
Flyout modal, in Safari, 'blinks' on left side before opening on the right


jeffchown:
### Flux version

v2.1.1

### Livewire version

v3.6.2

### Tailwind version

v.4.0.8

### Browser and Operating System

Safari 18.3 on MacOS

### What is the problem?

In a brand new Laravel 12 w/ Flux 2.0 app, if you try the Flux modal flyout example from https://fluxui.dev/components/modal#flyout, when clicking the `Edit profile` button, the modal will 'blink' on the left-hand side of the browser before opening on the right-hand side. e.g.

https://github.com/user-attachments/assets/25628487-3a97-45ff-8cf7-b19ec2b520ca

### Code snippets

Replace your `dashboard.blade.php` content with:

```
<x-layouts.app :title="__('Dashboard')">
    <flux:modal.trigger name="edit-profile">
        <flux:button>Edit profile</flux:button>
    </flux:modal.trigger>

    <flux:modal name="edit-profile" variant="flyout">
        <div class="space-y-6">
            <div>
                <flux:heading size="lg">Update profile</flux:heading>
                <flux:text class="mt-2">Make changes to your personal details.</flux:text>
            </div>

            <flux:input label="Name" placeholder="Your name" />

            <flux:input label="Date of birth" type="date" />

            <div class="flex">
                <flux:spacer />

                <flux:button type="submit" variant="primary">Save changes</flux:button>
            </div>
        </div>
    </flux:modal>
</x-layouts.app>
```

Then go to the app's `/dashboard` page and click `Edit profile` and you should see what is shown in the above video.

I narrowed it down to the following CSS block in `vendor/livewire/flux/dist/flux.css`:

```
@starting-style {
    [data-flux-modal] > dialog[open], [data-flux-modal] > dialog[open]::backdrop {
        opacity: 0;
    }

    [data-flux-modal] > dialog[open] {
        transform: scale(0.95);
    }

    [data-flux-modal] > dialog[open][data-flux-flyout] { transform: scale(1) var(--fx-flyout-translate, translateX(50px)); }
}
```
If you comment this out, then `npm run build` and try again, it works as expected.

According to https://developer.mozilla.org/en-US/docs/Web/CSS/@starting-style#browser_compatibility, @starting-style is compatible with Safari >= v17.5 and I am running 18.3, but something isn't quite right.

[...]

PR #1402, opened 2025-03-26
Fix flyout modal and toasts display bug

joshhanley:
This has a corresponding Flux Pro PR livewire/flux-pro#234

# The scenario

Currently if you open a flyout modal on Safari, the flyout initially appears on the left, and then jumps to the right instead of animating in. The same issue is also happening with toasts.

Recording 2025-03-26 at 18 22 02

```blade
<div> 
    <flux:modal.trigger name="edit-profile">
        <flux:button>Edit profile</flux:button>
    </flux:modal.trigger>

    <flux:modal name="edit-profile" variant="flyout">
        <div class="space-y-6">
            <div>
                <flux:heading size="lg">Update profile</flux:heading>
                <flux:text class="mt-2">Make changes to your personal details.</flux:text>
            </div>

            <flux:input label="Name" placeholder="Your name" />

            <flux:input label="Date of birth" type="date" />

            <div class="flex">
                <flux:spacer />

                <flux:button type="submit" variant="primary">Save changes</flux:button>
            </div>
        </div>
    </flux:modal>
</div>
```

# The problem

The issue is that to support RTL, we changed `ml-auto` to `ms-auto` and the issues have started appearing since then.

I'm not 100% on what is causing it, but I suspect it's because `ml-auto` becomes `margin-left: auto;` where as `ms-auto` becomes `margin-inline-start: auto;`, as such I suspect it's something to do with the inline part of it.

# The solution

The solution is to change back to using `ml-auto` and use `mr-auto` if RTL is detected. I also found that the fly in animation for flyouts and toasts were wrong on RTL, so fixed that too. Finally there was a duplicate `flyout` in the match so I have removed that.

```
ml-auto [--fx-flyout-translate:translateX(50px)] rtl:ml-0 rtl:mr-auto rtl:[--fx-flyout-translate:translateX(-50px)]
```

Recording 2025-03-26 at 18 32 15

Fixes livewire/flux#1375

### DIFF
--- a/dist/flux.css
+++ b/dist/flux.css
@@ -128,6 +128,14 @@ ui-toast {
                 transform: translateX(0.5rem);
             }
 
+            [dir="rtl"] &[data-position*="top"][data-position*="left"] {
+                transform: translateX(0.5rem);
+            }
+
+            [dir="rtl"] &[data-position*="top"][data-position*="right"] {
+                transform: translateX(-0.5rem);
+            }
+
             &[data-position*="bottom"] {
                 transform: translateY(0.5rem);
             }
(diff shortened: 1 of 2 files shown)


SRWieZ:
Thanks! I confirm having the same bug!

SRWieZ:
Just saw the same behaviors on toasts. You can test directly on https://fluxui.dev/components/toast

calebporzio:
ooof that's weird - killer fix josh thx

PR #1415, opened 2025-03-28
fix sidebar for RTL

3adeling:
# The scenario

Not sure why, but it seems that translations get ignored if rtl or ltr not specified for translation.

# The problem

in RTL, the sidemenu is opened to the right when the page loaded (disappear).

# The solution

I added both `ltr:` and `rtl:` to `translate-x-0!`

Fixes livewire/flux#

### DIFF
--- a/stubs/resources/views/flux/sidebar/index.blade.php
+++ b/stubs/resources/views/flux/sidebar/index.blade.php
@@ -22,7 +22,7 @@ if ($stashable) {
         'x-init' => '$el.classList.add(\'-translate-x-full\', \'rtl:translate-x-full\'); $el.removeAttribute(\'data-mobile-cloak\'); $el.classList.add(\'transition-transform\')',
     ])->class([
         'max-lg:data-mobile-cloak:hidden',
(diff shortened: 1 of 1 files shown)


joshhanley:
@3adeling thanks for the PR! Can you provide a screencapture of the issue you are seeing and add it to the scenario? That way Caleb can see what is going on? Thanks!

3adeling:
before:

image



After:

image

joshhanley:
@3adeling ah thanks!

This was actually an issue with Tailwind that got fixed in Tailwind v4.0.12.

I just tested with the Livewire starter kit and yeah out of the box it is broken. But if you upgrade Tailwind to v4.0.12 or later, you will find the issue is fixed without needing any changes to Flux (see screenshots below).

To do this run `npm update tailwindcss @tailwindcss/vite`.

I'm going to close this PR as it's not needed. Thanks again!

Default starter Tailwind version v4.0.8
image

After updating starter kit to Tailwind v4.0
image

itakwad:
@joshhanley  Thanyou

Issue #1416, opened 2025-03-28
Margin problem on input that has labels


altayevrim:
### Flux version

v2.1.1

### Livewire version

v3.6.2

### Tailwind version

v4.0.7

### Browser and Operating System

Safari on MacOS

### What is the problem?

Project has been created by Herd yesterday, it is a fresh one just for testing. Without Volt.

When I want to create an input with a table and add a mt-5 class. It applies that mt-5 to the div and not to the input which is expected behavior but the label itself is not inside that div hence the margin will be visible between the input and label and not between the label and other elements.
```
<hr />
<flux:input wire:model="sentence" label="Sentence" class="mt-5"/>
```
code looks like [this](https://shottr.cc/s/1zmc/SCR-20250328-k3l.png)
![preview](https://shottr.cc/s/1zmc/SCR-20250328-k3l.png)

### Code snippets

```
<hr />
<flux:input wire:model="sentence" label="Sentence" class="mt-5"/>
```

### How do you expect it to work?

The margin should be applied to the label not the input/input div.

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

jeffchown:
@altayevrim Whenever I'm working with the shorthand version of `flux:input`, if I want to apply a class, I always wrap it in a `div` first and apply the class to the div, e.g.:

```
<hr />
<div class="mt-5">
    <flux:input wire:model="sentence" label="Sentence" />
</div>
```

altayevrim:
That's make sense but when I see the Flux, it seems everything has been thought beforehand so that's why I wanted to report this in case it has been missed out. Otherwise, I'd definitely be using longhand as it's described [here](https://fluxui.dev/components/input#longhand)

Thanks for the tip though!

jeffchown:
You're welcome, @altayevrim .  When any component might expand into multiple markup elements (as the `flux:input` does), I find it best practice to wrap it in a div so I know what I'm applying the class to and there are no surprises that might break my layout.  

Helps keep things clean and predictable and doesn't count on the custom component to have to add too many conditionals to cover/consider all possibilities which could make the component's API overly complex.

joshhanley:
[...]

PR #1429, opened 2025-03-31
Enable scoped attributes for field components

joshhanley:
# The scenario

Currently if you have a input that has a label using the short hand syntax, if you try to add margin top to that component `class="mt-6"`, it actually puts the margin between the label and the input instead of above the input which is what is intended.

```blade
<div>
    <flux:heading size="xl">Profile</flux:heading>

    <flux:input class="mt-6" label="Name" />
</div>
```

image

# The problem

The issue is that any classes that are passed into the input component are being applied to the input components itself and not the wrapping field/label component.

The input component at the moment is structured like this (simplified for demonstration purposes)

```blade
<flux:with-field :$attributes :$name>
    <div {{ $attributes->only('class') }}>
    <input
            type="{{ $type }}"
            {{ $attributes->except('class')->class($classes) }}
        >
</flux:with-field>
```

# The solution

The solution proposed in this PR is to add support to the `with-field` component for the new component attribute prefixes, for example `field:class` or `label:class`.

The changes in this PR allow the following field components to be targeted:
- field - `field:class`
- label - `label:class`
- description - `description:class`
- error - `error:class`

This means that for the example at the top, we can change `class="mt-6` to `field:class="mt-6"` to ensure the margin top gets applied to the wrapping `field` component.

```blade
<div>
    <flux:heading size="xl">Profile</flux:heading>

    <flux:input field:class="mt-6" label="Name" />
</div>
```

image

Fixes livewire/flux#1416

### DIFF
--- a/stubs/resources/views/flux/with-field.blade.php
+++ b/stubs/resources/views/flux/with-field.blade.php
@@ -36,7 +36,6 @@ extract(Flux::forwardedAttributes($attributes, [
             <flux:description :attributes="Flux::attributesAfter('description:', $attributes, [])">{{ $descriptionTrailing }}</flux:description>
         <?php endif; ?>
     </flux:field>
-    <div {{ $attributes }}></div>
 <?php else: ?>
     {{ $slot }}
 <?php endif; ?>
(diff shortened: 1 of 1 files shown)


seabasss:
Sweet! Thanks @joshhanley 😀

jeffchown:
@joshhanley Nice!  I assume the normal, unscoped `class=` would still be applied the 'old' way so it's not a breaking change?

joshhanley:
@jeffchown yep, haven't changed any of that 🙂

PR #1432, opened 2025-03-31
Make callout container optional

inmanturbo:
# The scenario

# The problem

Screenshot from 2025-03-31 13-16-57

# The solution
Screenshot from 2025-03-31 13-23-29

Make callout container optional

### DIFF
--- a/stubs/resources/views/flux/callout/index.blade.php
+++ b/stubs/resources/views/flux/callout/index.blade.php
@@ -11,6 +11,7 @@
     'inline' => null,
     'text' => null,
     'icon' => null,
+    'container' => null,
 ])
 
 @php
@@ -25,7 +26,10 @@
     };
 
     $classes = Flux::classes()
-        ->add('@container p-2 flex border rounded-xl')
+        ->add(match(isset($container)) {
+            true => '@container p-2 flex border rounded-xl',
+            default => 'p-2 flex border rounded-xl'
+        })
         ->add([
             'border-(--callout-border) bg-(--callout-background)',
             '[&_[data-slot=heading]]:text-(--callout-heading)',


joshhanley:
@inmanturbo thanks for the PR! You haven't explained why this is needed. Can you please fill out our PR template properly otherwise we will need to close this PR. I've included it below. Thanks!

# The scenario

Describe as succinctly as possible what the bug/problem is. Be sure to include the steps to reproduce the bug, screenshots of the issue, and a self-contained Volt class component, which includes all Blade variable definitions, to reproduce the issue. 

# The problem

Describe here in detail what you found is wrong with the current code in this package. Add snippets of code here so the maintainers don't need to search the codebase for the issue.

# The solution

Describe here what you did to fix the problem, whether there were other possible solutions, and why you chose this one. Be sure to include snippets of the code changes you have made and screenshots of the problem fixed after your changes.

Fixes livewire/flux#

inmanturbo:
@joshhanley Thank you! it may not be needed, honestly. Just the callout gets squished when using `@container` if parent div is a using `flex-col` and there isn't enough sibling content in the div to otherwise expand it out enough to accommodate. I will fill out the template if/when I get time :smile: . Marking as draft till then. Easy enough to fix by publishing of you are on `dev-main`.

joshhanley:
@inmanturbo that'd be great thanks! Yep we like to see the code with the issue so we can test it locally ourselves to make sure it's the best solution before we merge it 🙂

inmanturbo:
@joshhanley of course! No worries. I'll mark it ready when It's done.

calebporzio:
closing due to inactivity - feel free to re-submit. thanks everyone!

PR #1434, opened 2025-04-01
Fix input leading icons appearing above headers

joshhanley:
# The scenario

Currently if you have an input with a leading icon and the page is scrolled, the leading icon is appearing above the header element. But this doesn't happen for trailing icons.

image

```blade
<flux:input icon="key" value="FLUX-1234-5678-ABCD-EFGH" readonly copyable />
```

# The problem

The issue is that input leading icons have a `z-10` on them, causing them to appear above the `<flux:header>` component.

# The solution

I'm not exactly sure why the z-index was on there, but this PR removes it. In my testing I couldn't find why that might have been required.

image


calebporzio:
let's try this - if it bites us we can use isolate on the root element of the thing containing the problem - thanks Josh

PR #1509, opened 2025-04-18
Automatically resolve Gravatar image by email

cord:
With new parameter Email (can be Email or hash) the avatar can show the gravatar image if available, otherwise fall back.

# The scenario

Make it easy to display gravatar image with fallback

# The problem

Currently we need to write some code to resolve a garavatar image source.

# The solution

some check and here we go

### DIFF
--- a/stubs/resources/views/flux/avatar/index.blade.php
+++ b/stubs/resources/views/flux/avatar/index.blade.php
@@ -8,6 +8,7 @@
     'color' => null,
     'badge' => null,
     'name' => null,
+    'email' => null,
     'icon' => null,
     'size' => 'md',
     'src' => null,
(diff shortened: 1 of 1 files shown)


joshhanley:
@cord thanks for the PR! I believe this is something that you should actually do in your user model in an accessor, not in the avatar component itself. The reason for this is not everyone will want to use Gravatar, so it's better to handle this on a project by project basis. Because of that I'm going to close this PR. Thanks for being willing to contribute!

cord:
@joshhanley got you point, maybe you could add a helper library as part of flux to support common cases like this?

joshhanley:
@cord thanks for the suggestion! Can you open that as a feature request https://github.com/livewire/flux/discussions/new?category=feature-requests? Thanks

PR #1622, opened 2025-05-21
Add color prop for primary buttons

joshhanley:
This PR adds support for adding a `color` prop to the `button` element for primary buttons. This will only work for buttons with `variant="primary"` as they are the ones that use accent colours.

image

```blade
<div>
    <flux:button variant="primary" color="blue">Primary</flux:button>
    <flux:button variant="primary" color="yellow">Primary</flux:button>
    <flux:button variant="primary" color="green">Primary</flux:button>
</div>
```

### DIFF
--- a/stubs/resources/views/flux/button/index.blade.php
+++ b/stubs/resources/views/flux/button/index.blade.php
@@ -18,12 +18,6 @@
 ])
 
 @php
-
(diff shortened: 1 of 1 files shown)


calebporzio:
jolly good matey

PR #1652, opened 2025-06-03
Fix misalignment when using badge in flux:label

rzv-me:
# The problem
There is a misalignment when using `<flux:label>` with a badge
CleanShot 2025-06-03 at 19 01 59@2x

# The solution

Simply adding inset top bottom, `-mb-1 -mt-1` fixes the problem

# Code to reproduce
```blade
<div class="space-y-6 md:space-y-0 md:grid grid-cols-2 gap-4">
    <flux:input label="Email" type="email"/>
    <flux:input badge="Optional" label="Phone" type="phone"/>
</div>
```

### DIFF
--- a/stubs/resources/views/flux/label.blade.php
+++ b/stubs/resources/views/flux/label.blade.php
@@ -15,7 +15,7 @@
     {{ $slot }}
 
     <?php if (is_string($badge)): ?>
-        <span class="ms-1.5 text-zinc-800/70 text-xs bg-zinc-800/5 px-1.5 py-1 rounded-[4px] dark:bg-white/10 dark:text-zinc-300 -mb-1 -mt-1" aria-hidden="true">
+        <span class="ms-1.5 text-zinc-800/70 text-xs bg-zinc-800/5 px-1.5 py-1 -my-1 rounded-[4px] dark:bg-white/10 dark:text-zinc-300" aria-hidden="true">
             {{ $badge }}
         </span>
     <?php elseif ($badge): ?>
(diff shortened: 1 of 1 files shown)


joshhanley (approved):
@rzv-me thanks for the PR! Changes look good to me. In future, please use our PR template as it is laid out. Thanks!

calebporzio:
Thanks! Can someone do a deeper dive on why this is needed? is it a line-height thing? just a regular height thing? is there something we should change to the badge component?

i'm fine going with this, but i just don't want to end up in a bind because we don't understand why this is a fix

joshhanley:
@calebporzio yeah so the line height of the label text and the badge text are both set to 16px. But the badge has `py-1` on it, which is causing it's height to be 24px.

So then when you have a label without a badge, next to a label with a badge, the badge/label combo is higher, which is then pushing it's input down.

I've refactored this PR to use `-my-1` and put it next to the `py-1`, to show it is offsetting the padding. It essentially makes the badge inset on the y-axis.

The badge used here is a custom implementation within the label component, it doesn't use the badge component, so it won't impact anything else.

```blade
<ui-label {{ $attributes->class($classes) }} data-flux-label>
    {{ $slot }}

    <?php if (is_string($badge)): ?>
        <span class="ms-1.5 text-zinc-800/70 text-xs bg-zinc-800/5 px-1.5 py-1 -my-1 rounded-[4px] dark:bg-white/10 dark:text-zinc-300" aria-hidden="true">
            {{ $badge }}
        </span>
    <?php elseif ($badge): ?>
        <span class="ms-1.5" aria-hidden="true">
            {{ $badge }}
        </span>
    <?php endif; ?>

    <?php if ($aside): ?>
        <span class="ms-1.5 text-zinc-800/70 text-xs bg-zinc-800/5 px-1.5 py-1 -my-1 rounded-[4px] dark:bg-white/10 dark:text-zinc-300" aria-hidden="true">
            {{ $aside }}
        </span>
    <?php endif; ?>
</ui-label>
```

calebporzio:
thanks

PR #1655, opened 2025-06-04
Fix: Sidebar visibility in RTL mode when stashable

noor4apps:
# The scenario

The `flux:sidebar` component, when configured with the `stashable` prop, fails to display correctly when the HTML page direction is set to `dir="rtl"`. Instead of sliding into view, the sidebar remains completely invisible on bigger screens. The component works as expected in LTR mode.

**Expected behavior:**
The sidebar should slide into view from the right side of the screen.

**Actual behavior:**
The sidebar does not appear; it remains completely invisible.

```
<html dir="rtl">
<body>
        <flux:sidebar sticky stashable class="border-e border-zinc-200 bg-zinc-50 dark:border-zinc-700 dark:bg-zinc-900">
                 //
        </flux:sidebar>
    </body>
</html>
```

# The problem
The current implementation within the index.blade.php for the stashable sidebar relies on adding both -translate-x-full and rtl:translate-x-full classes in the x-init directive:
```
if ($stashable) {
    $attributes = $attributes->merge([
        // ... other attributes ...
        'x-init' => '$el.classList.add(\'-translate-x-full\', \'rtl:translate-x-full\'); $el.removeAttribute(\'data-mobile-cloak\'); $el.classList.add(\'transition-transform\')',
    ])->class([
        // ... other classes ...
        '[[data-show-stashed-sidebar]_&]:translate-x-0! lg:translate-x-0!',
        // ...
    ]);
}
```
In RTL mode, Tailwind CSS specificity rules ensure that rtl:translate-x-full (which applies transform: translateX(100%)) overrides -translate-x-full (which applies transform: translateX(-100%)). This correctly sets the initial off-screen position to the right.

However, when the sidebar is toggled to be shown, the [[data-show-stashed-sidebar]_&]:translate-x-0! class (applying transform: translateX(0) !important) seems unable to correctly override or reset this initial transformation in RTL, leading to the sidebar remaining off-screen or otherwise invisible. The exact cause of this override failure with the current class setup in RTL is unclear but points to an interaction issue.

# The solution
The proposed solution is to make the initial off-screen transformation class more explicit within the x-init directive by detecting the document's direction using JavaScript and applying the correct Tailwind utility class directly. This removes the reliance on adding both LTR and RTL-specific transform classes simultaneously and letting CSS specificity sort it out, which might be the source of the conflict.

Code changes in index.blade.php:

The x-init attribute for the stashable sidebar is modified as follows:
```
if ($stashable) {
    $attributes = $attributes->merge([
        'x-bind:data-stashed' => '! screenLg',
        'x-resize.document' => 'screenLg = window.innerWidth >= 1024',
        'x-init' => "
            const isRtl = document.documentElement.dir === 'rtl';
            if (isRtl) {
                \$el.classList.add('translate-x-full'); // Hides sidebar to the right in RTL
            } else {
                \$el.classList.add('-translate-x-full'); // Hides sidebar to the left in LTR
            }
            \$el.removeAttribute('data-mobile-cloak');
            \$el.classList.add('transition-transform');
        ",
    ])->class([
        'max-lg:data-mobile-cloak:hidden',
        '[[data-show-stashed-sidebar]_&]:translate-x-0! lg:translate-x-0!',
        'z-20! data-stashed:start-0! data-stashed:fixed! data-stashed:top-0! data-stashed:min-h-dvh! data-stashed:max-h-dvh!'
    ]);
}
```
This modification avoids the conflict caused by using two classes simultaneously, and ensures that the correct transformation is applied based on the page orientation, which solves the problem of the gibberish not appearing in RTL mode.

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,8 @@
+# Default ignored files
+/shelf/
+/workspace.xml
+# Editor-based HTTP Client requests
+/httpRequests/
+# Datasource local storage ignored files
+/dataSources/
+/dataSources.local.xml
--- a/stubs/resources/views/flux/sidebar/index.blade.php
+++ b/stubs/resources/views/flux/sidebar/index.blade.php
@@ -19,7 +19,7 @@ if ($stashable) {
     $attributes = $attributes->merge([
         'x-bind:data-stashed' => '! screenLg',
         'x-resize.document' => 'screenLg = window.innerWidth >= 1024',
-        'x-init' => '$el.classList.add(\'-translate-x-full\', \'rtl:translate-x-full\'); $el.removeAttribute(\'data-mobile-cloak\'); $el.classList.add(\'transition-transform\')',
+        'x-init' => "const isRtl = document.documentElement.dir === 'rtl';if (isRtl) {\$el.classList.add('translate-x-full');} else {\$el.classList.add('-translate-x-full');}\$el.removeAttribute('data-mobile-cloak');\$el.classList.add('transition-transform');",
     ])->class([
         'max-lg:data-mobile-cloak:hidden',
         '[[data-show-stashed-sidebar]_&]:translate-x-0! lg:translate-x-0!',


noor4apps:
I'm checking out your components and would love to test drive the Pro version. I'm particularly focused on how they handle Right-to-Left (RTL) layouts for a project I'm on.

Any chance you could provide a Pro license key so I can give them a thorough review from an RTL perspective?

Thanks a lot!

rzv-me:
I tried to reproduce the issue described and I couldn't
but I think your solution is overkill, and just adding `rtl:lg:tranlate-x-0!` should be enough

can you add your PHP, Laravel, Flux and TailwindCSS version?

noor4apps:
PHP 8.2.28 
Laravel 12.17.0
Flux 2.1.6
Tailwindcss 4.0.7

joshhanley:
@noor4apps thanks for the PR and amazing write up! This issue was actually a Tailwind v4 issue not parsing classes in Blade properly. Please upgrade it to at least v4.0.12 or later. I always recommend ensuring everything is up to date first if you are experiencing issues. Because of this, I’m going to close the PR. If you’re still experiencing problems after an update, let us know and we can investigate further. Thanks for being willing to contribute!

noor4apps:
Confirming that upgrading Tailwind did indeed solve the problem.

joshhanley:
@noor4apps thanks for confirming!

PR #1690, opened 2025-06-14
Allow Variant to be Specified for Navbar Item Badge

JonPurvis:
# The scenario

When using the Navbar component and you want to add a Badge to a Navbar Item, you currently can't specify the variant of the badge you want to use, whilst you can change the colour, there's no variant option.

# The problem

The normal Badge component allows you to specify `variant="solid"`, which renders a bold, high contrast badge. The `navlist/badge.blade.php` doesn't have this option though, so you're left with a faded version.

# The solution

I've taken inspiration from how the normal Badge component handles the solid variant and made it possible, much like `badge-color`, to pass in `badge-variant`. Within the blade file, I've just copied the existing code from `badge.blade.php` as that works fine.

## Before
Screenshot 2025-06-14 at 19 04 59

## After
Screenshot 2025-06-14 at 19 03 10

### DIFF
--- a/stubs/resources/views/flux/navlist/badge.blade.php
+++ b/stubs/resources/views/flux/navlist/badge.blade.php
@@ -1,4 +1,5 @@
 @props([
+    'variant' => null,
     'color' => null,
 ])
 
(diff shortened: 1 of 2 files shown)


JonPurvis:
Hey @joshhanley 

Would there be any interest in accepting this? 🤞 

Thanks! 🍻

joshhanley:
@JonPurvis hey, that's up to Caleb and he's head down on Laracon US prep. So might be a bit before he gets to review it.

JonPurvis:
Hey, no worries! Thanks for the clarification 😄

calebporzio:
Will merge. @joshhanley - can you add this to the docs?  thanks everyone!

joshhanley:
I started to document this and realised this has only been applied to the navlist.item/navlist.badge and not navbar.item/navbar.badge. Also there is a bunch of duplicated classes and unused classes. So I've submitted a new PR to fix these up. See PR #1848.

JonPurvis:
Thanks @joshhanley and @calebporzio 🥳

Issue #1691, opened 2025-06-15
Sidebar Arrow Direction Issue in RTL Mode


EhsanGhafoori:
### Flux version

2

### Livewire version

3

### Tailwind version

4

### Browser and Operating System

chrome

### What is the problem?

Hi, I noticed a small issue with the sidebar arrows in RTL (right-to-left) mode. When the layout switches to RTL, the toggle/collapse arrows in the sidebar don't flip direction — they still point as if in LTR.

This makes the UX feel a bit off in RTL. Expected behavior: arrows should flip horizontally in RTL to reflect the change in direction.

Thanks for the amazing work on Flux UI 💜

![Image](https://github.com/user-attachments/assets/8124ce29-0d87-40a7-854d-832ac7bc307d)
![Image](https://github.com/user-attachments/assets/6c1566af-73b9-4f88-85f3-707979033950)

### Code snippets

        <flux:navlist.group expandable :expanded="request() -> routeIs('accounting.human-resources.*')" heading="{{ __('messages.Human Resources') }}">
            <flux:navlist.item icon="employees" href="{{ route('accounting.human-resources.employees.index') }}">
                {{ __('messages.Employees') }}
            </flux:navlist.item>
            <flux:navlist.item icon="roles" href="{{ route('accounting.human-resources.roles.index') }}">
                {{ __('messages.Roles') }}
            </flux:navlist.item>
        </flux:navlist.group>

### How do you expect it to work?

When the layout switches to RTL, the toggle/collapse arrows in the sidebar don't flip direction Which is shouldbe oposite

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

joshhanley:
@EhsanGhafoori thanks for reporting! I've submitted a PR with a fix.

PR #1693, opened 2025-06-15
Fix RTL navlist group closed chevron

joshhanley:
# The scenario

Currently if you are using the navlist group on an RTL site, the closed chevron is still pointing right instead of left.

image

# The problem

The issue is we have no handling for RTL in the navlist group component, so it needs to be added.

# The solution

The solution is to ensure we rotate the chevron to the left instead of the right when RTL is enabled.

I've done this by adding `rtl:rotate-180` to the icon.

image

Fixes livewire/flux#1691


calebporzio:
great, thanks

PR #1727, opened 2025-06-25
Fixed button padding

### LINKED ISSUE #1728: bug introduced with "Commit 8de57c8: adjust button padding with icons and subtle variant color" for <flux:button>
### Flux version

v2.2.1

### Livewire version

v3.6.3

### Tailwind version

v4.0.7

Chrome on Windows and Firefox on Windows

### What is the problem?

The commit "8de57c8" introduced a padding bug in `<flux:buttons>` with **normal size only** if there is a `wire:click` directive on it or if the button has `type="submit"`.

The problem is duo to the change in line 57 of the [‎stubs/resources/views/flux/button/index.blade.php](https://github.com/livewire/flux/commit/8de57c82ab8cdc9dc84ab03861062a7a463a380d#diff-e8effb2d215d2df7153847eebece7824f3aaa68b9b37d5c0ef759d2fd01f2c64).

The directive `wire:click` and the `type="submit"` add a `<div>` with a `<svg>` spinner  before the `<span>` that contain the text in the button.
This has the effect to apply the "ps-3" because the first-child is not a `<span>`.

The visual effect is the following:

Image

(the button element has the px-4 and ps-3 classes for padding).

But the expected visual effect should be:

Image

 (the button element has only px-4 class for padding).

### Code snippets

```blade
<flux:button wire:click="console.log('hi')">
 AI
</flux:button>
```

I expect that the "ps-3" and "pe-3" class should both or none of them be applied if the icon is not visible.

```blade
<flux:button>
 AI
</flux:button>
```

sertxudev:
This PR fixed a visual issue with the button padding if the button is type submit.

## Visual issue before:

```html
<flux:button variant="primary">Save</flux:button>
```
image

```html
<flux:button variant="primary" type="submit">Save</flux:button>
```
image

As you can see the left padding is smaller if the button is type submit, so the text is not centered.

Fixes #1728

### DIFF
--- a/stubs/resources/views/flux/button/index.blade.php
+++ b/stubs/resources/views/flux/button/index.blade.php
@@ -62,7 +62,7 @@ $classes = Flux::classes()
     ->add('relative items-center font-medium justify-center gap-2 whitespace-nowrap')
     ->add('disabled:opacity-75 dark:disabled:opacity-75 disabled:cursor-default disabled:pointer-events-none')
     ->add(match ($size) { // Size...
-        'base' => 'h-10 text-sm rounded-lg' . ' ' . ($square ? 'w-10' : 'px-4 [&:has(>:not(span):first-child)]:ps-3 [&:has(>:not(span):last-child)]:pe-3'),
+        'base' => 'h-10 text-sm rounded-lg' . ' ' . ($square ? 'w-10' : 'px-4 [&:has(>:not(span):not([data-flux-loading-indicator]):first-child)]:ps-3 [&:has(>:not(span):not([data-flux-loading-indicator]):last-child)]:pe-3'),
         'sm' => 'h-8 text-sm rounded-md' . ' ' . ($square ? 'w-8' : 'px-3'),
         'xs' => 'h-6 text-xs rounded-md' . ' ' . ($square ? 'w-6' : 'px-2'),
     })


joshhanley:
@sertxudev thanks for the PR! In future, please don't delete our PR template and instead fill it out. It's there for a reason 🙂

joshhanley:
# The scenario

Currently if you have a button which has a `wire:click` on it, the left padding is smaller than the right padding. See the screenshot below, where you can see the button without `wire:click` has even padding on both sides, but the one with `wire:click` is smaller on the left.

image

```blade
<?php

use Livewire\Component;

new class extends Component {
    //
} ?>

<div class="flex flex-col items-start gap-4">
    <div>
        <flux:button>Button</flux:button>
        - empty button
    </div>
    <div>
        <flux:button wire:click="example">Button</flux:button>
        - wire:click button
    </div>
</div>
```

# The problem

The issue is that the following code was added in 8de57c8

```
[&:has(>:not(span):first-child)]:ps-3 [&:has(>:not(span):last-child)]:pe-3
```

I believe that was added so that when icons are added, that it looks balanced. See screenshot below.

image

In the scenario where there is a `wire:click` on the button though, the first element inside of the button is actually the loading icon div.

```blade
<?php if ($loading): ?>
    <div class="absolute inset-0 flex items-center justify-center opacity-0" data-flux-loading-indicator>
        <flux:icon icon="loading" :variant="$iconVariant" :class="$iconClasses" />
    </div>
<?php endif; ?>
```

So what is happening is that the CSS selector is resolving to true, as the first element isn't a span, and so it's changing the left padding from 4 to 3.

But this shouldn't happen for the loading indicator and instead should happen only if there is a leading icon.

# The solution

The solution presented in this PR is to add a check to ensure that the first/last element is also not the loading indicator element.

```
[&:has(>:not(span):not([data-flux-loading-indicator]):first-child)]:ps-3
```

This fixes the problem for buttons without any icons. But now causes a misalignment between two buttons with icons if one has `wire:click` and the other doesn't.

image

With the plus symbol icon above, the button with the 3 padding left looks better.

But if you use a button that has an icon that takes up more visual space, then the alignment looks wrong with the 3 left padding.

image

So I'm wondering whether the solution is actually to just revert the additional classes instead? @calebporzio what do you think?

joshhanley:
Caleb didn't mean to merge this. I've pushed a fix e6a2ede that handles the different scenarios I presented:
- empty button
- `wire:click` button
- icon button
- `wire:click` button with icon

Now everything is consistent.

image

Issue #1737, opened 2025-06-27
Initials on profile vs avatar


mattcode72:
### Flux version

2.2.1

### Livewire version

3

### Tailwind version

4.1.10

### Browser and Operating System

Chrome on macOS

### What is the problem?

Initials on profile component is not managed in the same way as avatar component. For my name "Matteo", with avatar initials I get "Ma" and for profile I get "M".

### Code snippets

```blade
<flux:profile
                        avatar:name="{{ auth()->user()->name }}"
                        :name="auth()->user()->name"
                        avatar:color="auto"
                        icon-trailing="chevrons-up-down"
 />
```

### How do you expect it to work?

I want the props `:name="auth()->user()->name"` to give me two letters for a single name (Matteo for example) and not only one.

The code to change is : 

`// If no initials are provided, we'll try to generate them from the name by taking the first letter of the first and last name...
$initials ??= collect(explode(' ', $name ?? ''))
    ->map(fn($part) => Str::substr($part, 0, 1))
    ->filter()
    ->only([0, count(explode(' ', $name ?? '')) - 1])
    ->implode('');`

Thank you!

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

joshhanley:
@mattcode72 thanks for reporting! I've submitted a PR with a fix.

PR #1743, opened 2025-06-30
Fix profile initials

joshhanley:
# The scenario

Currently if you pass a name like "Caleb" to the avatar component, it will generate initials "Ca", but if you do the same for the profile component it generates initials "C".

image

```blade
<div>
    <flux:profile name="Caleb" />
    <flux:avatar name="Caleb" />
</div>
```

# The problem

The issue is that there is an inconsistency between how the initials are generated because the profile component has it's own initials generation logic which is different to the avatar logic.

# The solution

The profile component only passes the initials on to the avatar component and doesn't use them anywhere else, so there is no need for it to calculated them. So I've removed the initials logic from the profile component and it now falls through to the avatar component to calculate them.

Now both match and are consistent.

image

Fixes livewire/flux#1737

### DIFF
--- a/stubs/resources/views/flux/profile.blade.php
+++ b/stubs/resources/views/flux/profile.blade.php
@@ -14,13 +14,6 @@
 @php
 $iconTrailing = $iconTrailing ?? ($chevron ? 'chevron-down' : null);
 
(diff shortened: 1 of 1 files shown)


calebporzio:
great, yep thanks

Issue #1735, opened 2025-06-27
<flux:text> renders <div> instead of <p>


billpebert:
### Flux version

v2.1

### Livewire version

v2.1

### Tailwind version

v4.0.7

### Browser and Operating System

Chrome

### What is the problem?

<img width="891" alt="Image" src="https://github.com/user-attachments/assets/36260751-25a7-446a-acd6-4f7fd955a6f2" />

`<flux:text>` renders as `<div>` instead of a `<p>`

When i put the attribute _inline_, it renders as a`<span` as we expected in the [documentation](https://fluxui.dev/components/text#flux:text)

<img width="855" alt="Image" src="https://github.com/user-attachments/assets/c864475a-4041-49a0-b54c-ad64a2c131d6" />

The following is the packages in the composer.json

>     "require": {
        "php": "^8.2",
        "bacon/bacon-qr-code": "^3.0",
        "intervention/image": "^3.11",
        "laravel/framework": "^12.0",
        "laravel/nightwatch": "^1.7",
        "laravel/tinker": "^2.10.1",
        "livewire/flux": "^2.1",
        "livewire/flux-pro": "^2.1",
        "livewire/volt": "^1.7.0",
        "pragmarx/google2fa-laravel": "^2.3"
    },

### Code snippets

```blade                       
                        <flux:text>
                            This is flux:text
                        </flux:text>

                        <flux:text inline>
                            This is flux:text
                        </flux:text>```

### How do you expect it to work?

Expected the `<flux:text>` will render as `<p>`
```blade                       
                        <flux:text>
                            This is flux:text
                        </flux:text>```

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

joshhanley:
@billpebert thanks for reporting! I've submitted a PR with a fix.

PR #1744, opened 2025-06-30
Fix text component to use `<p>` tag

joshhanley:
# The scenario

Currently if you use the `<flux:text>` component it renders a `<div>` instead of a `<p>` tag. It's currently documented in the `inline` prop section of that docs that it is a `<p>` tag by default `If true, the text element will be a span instead of a p.`.

image

```blade
<flux:text>Some text</flux:text>
```

# The problem

The issue is that it is expected that it would render a `<p>` tag by default and render a `<span>` when it's inline.

But currently it outputs a `<div>`.

If you need a `<p>` tag, then you have to do the following which feels wrong.

```blade
<flux:text><p>Some text</p></flux:text>
```

# The solution

This PR changes it so it renders a `<p>` instead of a `<div>` which is more semantic anyway.

image

Fixes livewire/flux#1735

### DIFF
--- a/stubs/resources/views/flux/text.blade.php
+++ b/stubs/resources/views/flux/text.blade.php
@@ -39,4 +39,4 @@ $classes = Flux::classes()
     ;
 @endphp
 {{-- NOTE: It's important that this file has NO newline at the end of the file. --}}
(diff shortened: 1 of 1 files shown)


calebporzio:
this is good yep. thanks

Issue #1765, opened 2025-07-07
Navlist badge is not shown if content is falsy


mrextreme:
### Flux version

v2.2.1

### Livewire version

v3.6.3

### Tailwind version

v4.1.11

### Browser and Operating System

Chrome on macos

### What is the problem?

Navlist items' badge is not shown if the value is falsy, for example 0, 0.0 or "0", because the if statement in `vendor/livewire/flux/stubs/resources/views/flux/navlist/item.blade.php` line 86 simply says `<?php if ($badge): ?>`

### Code snippets

`<flux:navlist.item .... badge="0">Something</flux:navlist.item>`

### How do you expect it to work?

It should show 0 or 0.0.

Something like the below should do the trick.

```<?php if( $badge !== null && $badge !== '' ): ?>```
or
```<?php if( $badge !== '' ): ?>```

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

joshhanley:
@mrextreme thanks for reporting! I've submitted a PR with a fix.

PR #1769, opened 2025-07-08
Fix navbar and navlist falsey badge check

joshhanley:
# The scenario

Currently if you have a navbar item or navlist item with a badge, if the passed in badge value is "0" no badges are shown.

image

```blade
<?php

use Livewire\Component;

new class extends Component {
    
} ?>

<div>
    <flux:navbar>
        <flux:navbar.item href="#">Home</flux:navbar.item>
        <flux:navbar.item href="#" badge="0">Inbox</flux:navbar.item>
    </flux:navbar>
    <flux:navlist class="w-64">
        <flux:navlist.item href="#" icon="home">Home</flux:navlist.item>
        <flux:navlist.item href="#" icon="envelope" badge="0">Inbox</flux:navlist.item>
    </flux:navlist>
</div>
```

# The problem

The issue is that both components are only checking `<?php if ($badge): ?>` but "0" returns a falsey value, hence why nothing is being shown.

# The solution

The solution is to change both components to ensure that badge isset and not an empty string.

```blade
<?php if (isset($badge) && $badge !== ''): ?>
```

Now a "0" badge value gets displayed correctly.

image

Fixes livewire/flux#1765


calebporzio:
ah of course...good catch. you can merge after resolving conflicts

joshhanley:
Done!

PR #1817, opened 2025-07-25
Add functional keyboard shortcuts to Flux buttons

00bx:
## What this does

I don't know if that's something that already exists and I haven't made things more complicated, but I use this in any project I make on a daily basis. I wanted to share it with you.
it Adds working keyboard shortcuts to buttons. Flux already shows keyboard hints in tooltips, but this makes them actually **trigger the button action** when pressed.

With `kbd="⌘S"` you get:
- Visual hint: "Save ⌘S" displayed on the button
- **Functional shortcut**: Button actually clicks when Cmd+S is pressed

## Why we need this

Flux tooltips can show `kbd="⌘S"` hints, but the shortcuts don't work. To make them functional, you have to:
1. Write custom Alpine.js code for each button
2. Convert ⌘S symbols into "cmd.s" Alpine format manually
3. Connect keyboard events to button actions yourself

This is tedious and inconsistent across projects.

## How it works

Just add the `kbd` prop to any button:

```php
<flux:button kbd="⌘S" wire:click="save">Save</flux:button>
```

Now when users press Cmd+S, the button automatically clicks and triggers your `wire:click="save"` action.

The shortcut also shows visually on the button so users know it exists.

## More examples

```php
<!-- Simple shortcuts -->
<flux:button kbd="⌘B" wire:click="bold">Bold</flux:button>
<flux:button kbd="⌘I" wire:click="italic">Italic</flux:button>

<!-- Multiple keys -->
<flux:button kbd="⌘⇧Z" wire:click="redo">Redo</flux:button>

<!-- Function keys -->
<flux:button kbd="F5" wire:click="refresh">Refresh</flux:button>
```

## What keys work

- **Mac symbols**: ⌘ ⇧ ⌥ ⌃ (Command, Shift, Option, Control)
- **Special keys**: F1-F12, Enter, Escape, Tab, Arrows
- **Regular keys**: A-Z, 0-9

## Technical details

Added a `ModifierParser` class that converts symbols like ⌘S into Alpine.js format (cmd.s). The button component now wraps itself in a div with Alpine.js keyboard listeners when a `kbd` prop is provided.

## Benefits

- **Easy to use**: Just add one prop
- **Shows shortcuts**: Users can see what keys to press
- **No breaking changes**: Completely optional
- **Works everywhere**: Any button, any action

thanks.

### DIFF
--- a/ModifierParser.php
+++ b/ModifierParser.php
@@ -1,0 +1,63 @@
+<?php
+
+namespace App\Actions;
+
+class ModifierParser
+{
+    /**
+     * Create a new class instance.
+     */
+    public function __invoke($kbd)
+    {
+        $modifiers = [
+            '⌘' => 'cmd',      // Command ( key)
+            '⇧' => 'shift',     // Shift
+            '⌥' => 'alt',       // Option/Alt
+            '⌃' => 'ctrl',      // Control
+            '⌫' => 'backspace', // Delete (backward)
+            '⌦' => 'delete',    // Forward Delete
+            '⎋' => 'escape',    // Escape
+            '↩' => 'enter',     // Return/Enter
+            '⇥' => 'tab',       // Tab
+            '␣' => 'space',     // Spacebar
+            '↑' => 'up',        // Up Arrow
+            '↓' => 'down',      // Down Arrow
+            '←' => 'left',      // Left Arrow
+            '→' => 'right',     // Right Arrow
+            '⇞' => 'pageup',    // Page Up
+            '⇟' => 'pagedown',  // Page Down
+            '↖' => 'home',      // Home
+            '↘' => 'end',       // End
+            '' => 'capslock',  // Caps Lock
+            '' => 'fn',        // Function key
+            'F1' => 'f1',
+            'F2' => 'f2',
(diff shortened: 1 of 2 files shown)


PrinceMinky:
Love this!

calebporzio:
Thank you for this contribution, however i have a few hangups:
- i think this might be a breaking change for folks that previously used kbd for display only and implemented a listener somewhere else
- this significantly bloats the already bloated button component
- might add slight runtime overhead but I'm not AS worried about that
- people might try other combinations that we may not have accounted for

although i definitely see the value and convenience here i think I'm going to say "no for now" - thank for your willingness to contribute!

zemenbeg:
> Thank you for this contribution, however i have a few hangups:
> 
> * i think this might be a breaking change for folks that previously used kbd for display only and implemented a listener somewhere else
> * this significantly bloats the already bloated button component
> * might add slight runtime overhead but I'm not AS worried about that
> * people might try other combinations that we may not have accounted for
> 
> although i definitely see the value and convenience here i think I'm going to say "no for now" - thank for your willingness to contribute!

Totally fine, i hope in future you guys can implement in even better way! thanks.

PR #1828, opened 2025-07-31
Update DateRangePreset.php

zhivkomonev:
adding last 12 months.

# The scenario

# The problem

# The solution

Fixes livewire/flux#


joshhanley:
@zhivkomonev thanks for the PR! In future please actually fill out our template as it is there for a reason and please don't open two PRs for the same thing. Both of these changes should have been in the same PR. But I'm going to close this as this PR won't work on it's own as it will require changes in Flux Pro as well. We are hoping to add better flexibility for date presets in the future.

Issue #1841, opened 2025-08-07
Not showing error messages from a named errorbag


josjojo:
### Flux version

v2.2.3

### Livewire version

v3.6.4

### Tailwind version

v4.1.11

### Browser and Operating System

Firefox on Windows

### What is the problem?

`<flux:input>` and `<flux:error>` will not show error messages from a named errorbag.

### Code snippets to replicate the problem

This is the PasswordController:
```php
class PasswordController extends Controller
{
    /**
     * Update the user's password.
     */
    public function update(Request $request): RedirectResponse
    {
        $validated = $request->validateWithBag('updatePassword', [
            'current_password' => ['required', 'current_password'],
            'password' => ['required', Password::defaults(), 'confirmed'],
        ]);

        $request->user()->update([
            'password' => Hash::make($validated['password']),
        ]);

        return back()->with('status', 'password-updated');
    }
}
```

This is the Blade code:
```blade
<div>
    <flux:input :label="__('Current Password')" name="current_password" type="password" autocomplete="current-password" />
    <flux:error name="current_password" />
</div>
```

### Screenshots/ screen recordings of the problem

When I have empty input fields and I click 'Save' I don't see any errors.
<img width="1162" height="485" alt="Image" src="https://github.com/user-attachments/assets/39751039-26f1-4b3e-a734-7eb815584570" />

### How do you expect it to work?

<img width="1156" height="557" alt="Image" src="https://github.com/user-attachments/assets/e0a8e836-3bfd-49fc-af2a-42ec37d409f8" />

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

josjojo:
I have also found a solution:
## Update `<flux:error>`
I copied the error blade template to `resources/views/flux/error.blade.php` and added `'errorbag' => 'default',` to the `@props`-array and changed
```php
$message ??= $name ? $errors->first($name) : null;
```
to
```php
$errorbag_obj = $errors->getBag($errorbag);
$message ??= $name ? $errorbag_obj->first($name) : null;
```

## Update `<flux:input>`
I copied the input blade template to `resources/views/flux/input/index.blade.php` and added `'errorbag' => 'default',` to the `@props`-array and changed
```php
$invalid ??= ($name && $errors->has($name));
```
to
```php
[...]

PR #1869, opened 2025-08-18
Add named error bag support to error component

joshhanley:
# The scenario

Currently if you are using Laravel's named error bag feature, the error component can't show errors from a named error bag.

```blade
@php
    $errorBag = $errors->getBag('other');
    $errorBag->add('name', 'Other Bag: Name is required');

    $errors->put('other', $errorBag);
@endphp

<flux:error name="name" />
```

# The problem

The issue is that the error component doesn't support named error bags. It accesses error messages by calling `$errors->first($name)` which will always resolve from the `default` error bag.

# The solution

This PR adds a `bag` prop to the error component so a named error bag can be specified.

```blade
<flux:error name="name" bag="other" />
```

I originally had `errorBag`, but I didn't like the multiword, so changed it to just `bag`.

Fixes livewire/flux#1841

### DIFF
--- a/stubs/resources/views/flux/error.blade.php
+++ b/stubs/resources/views/flux/error.blade.php
@@ -2,13 +2,15 @@
     'name' => null,
     'message' => null,
     'nested' => true,
+    'bag' => 'default',
 ])
 
 @php
-$message ??= $name ? $errors->first($name) : null;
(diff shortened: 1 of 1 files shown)


calebporzio:
I like it!

PR #1888, opened 2025-08-25
Remove whitespace-nowrap from navlist.item to allow wrapping on mobile

cstisa:
# The scenario

Using the <flux:sidebar stackable> component and having <flux:navlist.item> with a long text.

# The problem

image

The text does an overflow instead of being wrapped on small screens. And the way the component was created prevented me to override the tailwind class.

# The solution

image

I had to publish the component (navlist.item) and edit manually the tailwind class. It worked well after the update.

Fixes livewire/flux#

### DIFF
--- a/stubs/resources/views/flux/navlist/item.blade.php
+++ b/stubs/resources/views/flux/navlist/item.blade.php
@@ -74,7 +74,7 @@ $classes = Flux::classes()
     <?php endif; ?>
 
     <?php if ($slot->isNotEmpty()): ?>
-        <div class="flex-1 text-sm font-medium leading-none whitespace-nowrap [[data-nav-footer]_&]:hidden [[data-nav-sidebar]_[data-nav-footer]_&]:block" data-content>{{ $slot }}</div>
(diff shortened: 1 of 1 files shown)


calebporzio:
What do you. think about this @joshhanley

joshhanley:
@cstisa thanks for the PR! I've been testing this change out and while I understand the need for it, I doesn't look the greatest. See below for my test on our demo sidebar. This will require additional work to allow more spacing between the lines of the text, which then causes other layout issues with the hover and active states.

Screenshot 2025-09-05 at 08 15 08AM@2x

So for now I think we're going to close this PR and wait to see if others people run into this issue. But if you want this in your app for now, you can either publish the component and change it like you did, or if you'd prefer not to publish the component, you can add the class `class="[&>[data-content]]:whitespace-normal"`, like this:

```blade
<flux:sidebar.item href="#" class="[&>[data-content]]:whitespace-normal">Marketing site with some longer text and details</flux:sidebar.item>
```

Thanks for being willing to contribute!

Issue #1452, opened 2025-04-04
Mask for currency


junn-mendoza:
### Flux version

 v1.0.29.

### Livewire version

v3.4

### Tailwind version

v3.4.17

### Browser and Operating System

Chrome

### What is the problem?

just like to ask how to mask the currency in an flux:input 

### Code snippets

<flux:input wire:model.live="make" label="In year 3, I am to make a net profit of" placeholder="eg. £100,000"/>

### How do you expect it to work?

£ 1,000,000? £ 100,000?

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

jeffchown:
@junn-mendoza Best way is to use Alpine JS's https://alpinejs.dev/plugins/mask#money-inputs

joshhanley:
@junn-mendoza yep @jeffchown is correct, just use Alpine's mask for that. Going to close this issue.

PR #1894, opened 2025-08-26
Allow using mask:dynamic when using masks on input

marcorivm:
# The scenario
The FluxUI input component currently supports Alpine.js mask functionality through the `mask` attribute, which maps to `x-mask`. However, it doesn't support Alpine.js's dynamic masking feature (`x-mask:dynamic`), which allows masks to change based on user input. This is particularly useful for:
- Credit card inputs that need different formats for AMEX vs other cards
- Phone numbers with optional extensions
- Dynamic money formatting with the `$money()` function

Steps to reproduce the limitation:
```blade
{{-- This works (static mask) --}}
<flux:input mask="(999) 999-9999" />

{{-- This doesn't work (dynamic mask) --}}
<flux:input x-mask:dynamic="$money($input)" />
```

# The problem
The current implementation in `/stubs/resources/views/flux/input/index.blade.php` only handles static masks:

```blade
@props([
    // ... other props
    'mask' => null,
    // ... other props
])

// Line 140-141 - Only supports static x-mask
@if ($mask) x-mask="{{ $mask }}" @endif
```

This prevents developers from using Alpine.js's powerful dynamic masking features documented at https://alpinejs.dev/plugins/mask#mask-functions

# The solution
Added support for `mask:dynamic` by:

1. **Extracting the `mask:dynamic` attribute** using the existing FluxUI pattern (similar to `icon:trailing`):
```blade
@php $maskDynamic ??= $attributes->pluck('mask:dynamic'); @endphp
```

2. **Adding a `maskDynamic` prop** for flexibility:
```blade
@props([
    // ... other props
    'maskDynamic' => null,
    'mask' => null,
    // ... other props
])
```

3. **Updating the input element** to handle both static and dynamic masks, with dynamic taking precedence:
```blade
@if ($maskDynamic) x-mask:dynamic="{{ $maskDynamic }}" @elseif ($mask) x-mask="{{ $mask }}" @endif
```

## Usage Examples
This solution maintains full backward compatibility while enabling dynamic masks through the idiomatic FluxUI syntax:

```blade
{{-- Credit card with dynamic formatting --}}
<flux:input 
    mask:dynamic="$input.startsWith('34') || $input.startsWith('37') ? '9999 999999 99999' : '9999 9999 9999 9999'"
    placeholder="Credit card number" 
/>

{{-- Money formatting --}}
<flux:input 
    mask:dynamic="$money($input)"
    placeholder="Enter amount" 
/>

{{-- European money format --}}
<flux:input 
    mask:dynamic="$money($input, ',', '.', 2)"
    placeholder="Amount (€1.234,56)" 
/>

{{-- Phone with extension --}}
<flux:input 
    mask:dynamic="$input.length > 10 ? '999-999-9999 x99999' : '999-999-9999'"
    placeholder="Phone with extension" 
/>

{{-- Existing static mask still works --}}
<flux:input mask="(999) 999-9999" placeholder="Phone" />
```

## Alternative Approaches Considered
Skipping the usage of FluxUI when needing to use mask:dynamic but this makes it harder to mantain consistency on projects and needs the user to remember they need to use a different component in a very specific scenario.

Fixes livewire/flux#1452


marcorivm:
Hi @joshhanley yes, so x-mask:dynamic doesn't seem to work since it's not expected as an attribute on the current input component, it's only expecting the attribute "mask". 

I can look into leaving only the mask:dynamic syntax and remove the maskDynamic prop, but based on how the component already handles other similar properties; see iconTrailing, iconLeading and iconVariant, I think it would be more consistent to leave both sytnaxes available. What do you think?

joshhanley:
@marcorivm I just tested this, and it works fine for me:

```blade
<flux:input x-mask:dynamic="$money($input)" />
```

Screenshot 2025-09-03 at 06 31 46PM

In regards to the syntax, we are moving to `icon:trailing`, etc. syntax, which is why I requested the change.

marcorivm:
@joshhanley I'm sorry, I just tested using x-mask:dynamic and it worked. I'm not sure what might have been happening when I tried doing it before, but I couldn't replicate the issue anymore. 

I updated the PR to use only mask:dynamic, this will enable the use of mas:dynamic without the x-, while I think this is nice since makes applying a dynamic mask syntax the same as using a regular mask wthout having to think about alpinejs I also would understand if you don't want to merge this and want to keep it just using x-mask:dynamic

calebporzio:
thanks for the update. please add `'maskDynamic'` as a prop as well as the `mask:dynamic` forward then @joshhanley can merge

calebporzio:
Also should have docs for this as well @joshhanley

joshhanley:
Ha sorry @marcorivm! I will add the prop back in and merge it.

joshhanley (approved):

joshhanley:
Re-added the prop and added docs.

R00118189:
how do we make the mask:dynamic="$money($input)" app locale aware?

Issue #1887, opened 2025-08-25
Stacked toast get blip on dark mode when doing wire navigate.


arifbudimanar:
### Flux version

v2.2.5

### Livewire version

v3.6.4

### Tailwind version

v4.1.11

### Browser and Operating System

Edge on Windows

### What is the problem?

When using persisted stack toasts, the toast briefly flashes into light mode during navigation. This occurs whether the stack toast is expanded or collapsed. The issue does not occur with non-stack toasts.

### Code snippets to replicate the problem

```blade
@persist('toast')
    <flux:toast.group expanded>
        <flux:toast />
    </flux:toast.group>
@endpersist
```

### Screenshots/ screen recordings of the problem

https://github.com/user-attachments/assets/e5c2fa9f-f2e3-4943-aa69-36f36904440e

### How do you expect it to work?

It should not flashes into light mode during navigation.

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

joshhanley:
@arifbudimanar thanks for reporting! I've submitted a fix and it will be available in the next release.

arifbudimanar:
@joshhanley Thanks

Issue #1842, opened 2025-08-08
Flash of Light Mode When Navigating with wire:navigate and Editor Component


fernandust:
### Flux version

2.2.3

### Livewire version

3.6.4

### Tailwind version

4.1.11

### Browser and Operating System

Chrome and Mac OS

### What is the problem?

When using the editor component from FLUX UI in combination with Livewire’s wire:navigate, a visual flash occurs on subsequent navigations. The entire page momentarily switches to light mode before returning to dark mode.

### Code snippets to replicate the problem

Steps to Reproduce:
	1.	Start in dark mode.
	2.	Visit any page other than the one using the editor component.
	3.	Use wire:navigate to navigate to a page that includes the editor component.
	4.	Observe a flash where the page briefly switches to light mode, then returns to dark mode.

🔍 Note: This issue is reproducible directly on the FLUX UI documentation site. Start on a non-editor page, refresh, and then navigate to the page that contains the editor. You’ll notice the flash when the editor loads in dark mode.

### Screenshots/ screen recordings of the problem

https://github.com/user-attachments/assets/ae2e5ada-a0ed-41ef-84ec-11e43975582e

### How do you expect it to work?

Dark mode should be preserved without any flashing or momentary switch to light mode when navigating between pages.

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

joelstein:
Same issue here. If I watch the `<html>` tag in the browser console as I switch between pages, I can see the class property is updated each time. But when navigating to a page with `<flux:editor>`, the class property is completely removed and re-added, and the light mode flickers on/off. Or, perhaps it does this on all of the pages, but it's so fast as to be not-noticeable, but it is noticeable while the editor is loading.

slava-vishnyakov:
Here's a workaround if it helps:

```
<html @class(['dark' => request()->cookie('theme') === 'dark'])>
<head>
    <script>
        document.addEventListener('DOMContentLoaded', () => {
            function setThemeCookie() {
                const theme = document.documentElement.classList.contains('dark') ? 'dark' : 'light';
                document.cookie = `theme=${theme}; path=/; max-age=31536000; SameSite=Lax`;
            }
            setThemeCookie();
[...]

Issue #1501, opened 2025-04-16
Dark Mode Flash


mute92:
### Flux version

2.1.4

### Livewire version

3.6.3

### Tailwind version

4.1.4

### Browser and Operating System

Chrome

### What is the problem?

The page is flashing in dark mode. I am using Laravel 12 without starter kits.

https://github.com/user-attachments/assets/33494063-7295-4bf2-9810-d094a18557e3

### Code snippets

layout
```blade<!DOCTYPE html>
<html lang="{{ str_replace('_', '-', app()->getLocale()) }}">
    <head>
        <meta charset="utf-8">
        <meta name="viewport" content="width=device-width, initial-scale=1.0">

        <title>{{ $title ?? 'Page Title' }}</title>

         <!-- Fonts -->
        <link rel="preconnect" href="https://fonts.bunny.net">
        <link href="https://fonts.bunny.net/css?family=figtree:400,500,600&display=swap" rel="stylesheet" />

  @vite(['resources/css/app.css', 'resources/js/app.js'])

     @fluxAppearance
      @livewireStyles
</head>
<body class="min-h-screen bg-white dark:bg-zinc-800">
    <flux:sidebar sticky stashable class="bg-zinc-50 dark:bg-zinc-900 border-r rtl:border-r-0 rtl:border-l border-zinc-200 dark:border-zinc-700">
        <flux:sidebar.toggle class="lg:hidden" icon="x-mark" />
        <flux:brand href="#" logo="https://fluxui.dev/img/demo/logo.png" name="Acme Inc." class="px-2 dark:hidden" />
        <flux:brand href="#" logo="https://fluxui.dev/img/demo/dark-mode-logo.png" name="Acme Inc." class="px-2 hidden dark:flex" />
        <flux:input as="button" variant="filled" placeholder="Search..." icon="magnifying-glass" />
        <flux:navlist variant="outline">
            <flux:navlist.item icon="home" :href="route('dashboard')" :current="request()->routeIs('dashboard')">Dashboard</flux:navlist.item>
            <flux:navlist.item icon="user" badge="12" href="{{route('drivers')}}" :current="request()->routeIs('drivers')">Drivers</flux:navlist.item>
            <flux:navlist.item icon="document-text"  href="{{route('orders')}}">Orders</flux:navlist.item>
           
        </flux:navlist>
        <flux:spacer />
        <flux:navlist variant="outline">
[...]

PR #1897, opened 2025-08-27
Fix toast stack dark mode navigate flash

joshhanley:
# The scenario

Currently if you have a toast stack inside `@persist` with dark mode enabled, and you navigate to a new page, the toast will flash white for a moment before during dark again.

Screenshot 2025-08-27 at 11 45 39AM

```blade
@persist('toast')
    <flux:toast.group>
        <flux:toast /> 
    </flux:toast.group>
@endpersist
```

```blade
<div>
    <flux:button x-on:click="window.Flux.toast({
        heading: 'Changes saved',
        text: 'Your changes have been saved.',
        variant: 'success',
        duration: 0,
    })">
        Dispatch toast
    </flux:button>
    <flux:link href="/toasts-flash" class="block! mt-4" wire:navigate>Navigate to toasts page</flux:link>
</div>
```

# The problem

This issue sent me down a rabbit hole of navigate requests and dark mode flashes which there is a deeper issue in how Livewire and Flux handle re-applying dark mode after a navigate (see #1501 and #1842). But I was finding, for the most part, most elements would retain their dark styling between navigates if the navigate happened quickly enough.

But where as for toasts in a stack, no matter what I changed, they would always flash white.

As it turns out, the issue is that we are transitioning `all` on the toast dialog element as that is causing the colours to also transition. So in the moment when navigate applies the new dom and Flux hasn't re-added `dark` to the html element, the toast is white and then when `dark` is added, it's transitioning to dark styles, hence why we notice the flash.

```css
[data-flux-toast-dialog] {
     transition: all 0.35s allow-discrete;
}
```

# The solution

The solution is to be more specific with the transition CSS styles to ensure we're only transitioning the properties that we actually need.

```css
[data-flux-toast-dialog] {
    transition: opacity 0.35s allow-discrete, transform 0.35s allow-discrete, height 0.35s allow-discrete;
}
```

Now there is no flashing for toasts on navigate.

Screenshot 2025-08-27 at 11 53 51AM

Fixes livewire/flux#1887

### DIFF
--- a/dist/flux.css
+++ b/dist/flux.css
@@ -132,12 +132,12 @@ ui-toast {
     [data-flux-toast-dialog] {
         opacity: 0;
         transform: translate(0);
-        transition: all 0.35s allow-discrete;
+        transition: opacity 0.35s allow-discrete, transform 0.35s allow-discrete, height 0.35s allow-discrete;
 
         &.showing {
             opacity: 1;
             transform: translate(0);
-            transition: all 0.2s allow-discrete;
+            transition: opacity 0.2s allow-discrete, transform 0.2s allow-discrete, height 0.2s allow-discrete;
         }
     }
 
--- a/dist/flux.css
+++ b/dist/flux.css
@@ -211,7 +211,7 @@ ui-toast-group {
     [data-flux-toast-dialog] {
         position: absolute;
         transform: translate(0);
-        transition: all 0.35s allow-discrete;
+        transition: opacity 0.35s allow-discrete, transform 0.35s allow-discrete, height 0.35s allow-discrete;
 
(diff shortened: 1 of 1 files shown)


fernandust:
Hi Josh! Great work finding out the problem 👍 could we also patch the editor to prevent the flash happening there? That would be awesome since im going to production next week and I use the editor in a couple of places. Thanks! See #1842

joshhanley:
@fernandust thanks! Yep I was investigating that yesterday. I've got a fair idea of what is going on. Caleb and I are catching up on Friday to talk about it as it will require some changes in `wire:navigate`. No promises for next week, but we're definitely working on it.

PR #1902, opened 2025-08-28
Add `@pure` annotation to Lucide IconCommand stub

ThomasEnssner:
When creating a lucide icon using the artisan command, the `@pure` directive was missing.


calebporzio:
Oooh good call

Issue #1914, opened 2025-09-04
Tooltip in new sidebar item being improperly set


scottzirkel:
### Flux version

v2.3.0

### Livewire version

v3.6.4

### Tailwind version

v4.1.12

### Browser and Operating System

FF on macOS

### What is the problem?

The slot for the sidebar item tooltip is always used, despite the value being passed in via attributes.

### Code snippets to replicate the problem

From the collapsible sidebar demo:

```blade
<flux:sidebar.item icon="home" href="#" current>Home</flux:sidebar.item>
```
The tooltip here will be Home, as pulled from the component slot. 

```blade
<flux:sidebar.item icon="home" href="#" tooltip="Dashboard" current>Home</flux:sidebar.item>
```
The tooltip here will still be Home, as `$tooltip` is overwritten by `$slot` in the component.

### Screenshots/ screen recordings of the problem

<img width="1583" height="163" alt="Image" src="https://github.com/user-attachments/assets/363e49ac-4f51-47f7-80b4-59b5d54b828b" />

<img width="793" height="57" alt="Image" src="https://github.com/user-attachments/assets/a2fde968-d2cd-462d-8ce8-7d5711142b82" />

### How do you expect it to work?

I would expect the tooltip I provided to appear, not the content from `$slot`

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

PR #1915, opened 2025-09-04
Reassigns tooltip

scottzirkel:
# The scenario

From the collapsible sidebar demo:

```blade
<flux:sidebar.item icon="home" href="#" current>Home</flux:sidebar.item>
```
The tooltip here will be Home, as pulled from the component slot. 

```blade
<flux:sidebar.item icon="home" href="#" tooltip="Dashboard" current>Home</flux:sidebar.item>
```
The tooltip here will still be Home, as `$tooltip` is overwritten by `$slot` in the component.

# The problem

```blade
$tooltip = $tooltip ?? $slot->isNotEmpty() ? (string) $slot : null;
```

While this looks right, it's actually setting the `$slot` value to `$tooltip`.

# The solution

```blade
$tooltip ??= $slot->isNotEmpty() ? (string) $slot : null;
```

Shifting to a slightly modified syntax solves the issue.

Example screenshot provided from simplified tinkerwell code, with result.

Image

Fixes livewire/flux#1914

### DIFF
--- a/stubs/resources/views/flux/sidebar/item.blade.php
+++ b/stubs/resources/views/flux/sidebar/item.blade.php
@@ -18,7 +18,7 @@
 ])
 
 @php
-$tooltip = $tooltip ?? $slot->isNotEmpty() ? (string) $slot : null;
+$tooltip ??= $slot->isNotEmpty() ? (string) $slot : null;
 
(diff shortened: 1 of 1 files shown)


joshhanley:
@scottzirkel looks good, thanks for the PR!

Issue #1936, opened 2025-09-09
Sidebar Group Heading on level 2


slovenianGooner:
### Flux version

v2.2.3

### Livewire version

v3.6.4

### Tailwind version

v4.0.8

### Browser and Operating System

Chrome on MacOS

### What is the problem?

Heading of a sidebar group is not visible if on level 2. The heading is hidden correctly for first-level in the collapsed sidebar, but any subsequent level the heading is hidden, when in reality it should display virtually the same as a sidebar item.

### Code snippets to replicate the problem

```blade
<flux:sidebar sticky collapsible
              class="bg-zinc-50 dark:bg-zinc-900 border-r border-zinc-200 dark:border-zinc-700">
    <flux:sidebar.header>
        <flux:sidebar.brand href="{{ route('dashboard') }}" :name="config('app.name')"
                            class="[&>div]:text-sm">
            <x-slot:logo>
                <x-logo.square class="text-zinc-900 dark:text-white size-6"/>
            </x-slot:logo>
        </flux:sidebar.brand>
        <flux:sidebar.collapse/>
    </flux:sidebar.header>

    <flux:sidebar.nav>
        <flux:sidebar.item icon="home" href="#" current>Home</flux:sidebar.item>
        <flux:sidebar.item icon="inbox" badge="12" href="#">Inbox</flux:sidebar.item>
        <flux:sidebar.item icon="document-text" href="#">Documents</flux:sidebar.item>
        <flux:sidebar.item icon="calendar" href="#">Calendar</flux:sidebar.item>
        <flux:sidebar.group icon="star" expandable heading="Favorites" class="grid">
            <flux:sidebar.item href="#">Marketing site</flux:sidebar.item>
            <flux:sidebar.item href="#">Android app</flux:sidebar.item>
            <flux:sidebar.item href="#">Brand guidelines</flux:sidebar.item>
            <flux:sidebar.group icon="arrow-right" expandable heading="Other" class="grid">
                <flux:sidebar.item href="#">Marketing site</flux:sidebar.item>
                <flux:sidebar.item href="#">Android app</flux:sidebar.item>
                <flux:sidebar.item href="#">Brand guidelines</flux:sidebar.item>
            </flux:sidebar.group>
        </flux:sidebar.group>
    </flux:sidebar.nav>
</flux:sidebar>
```

### Screenshots/ screen recordings of the problem

[...]

PR #1949, opened 2025-09-15
Fix sidebar nested groups in icon collapsed view

joshhanley:
# The scenario

Currently if you have a sidebar group inside another sidebar group, when the sidebar is collapsed down to icons, the nested group is missing it's heading and only showing icon in the dropdown.

Screenshot 2025-09-15 at 02 59 13PM@2x

# The problem

The issue is that we detect if the group is inside of a icon collapsed sidebar and if it is, we don't show the heading.

But if the group is within another group, then it should be shown.

# The solution

The solution is to add a check to see if the sidebar group is actually inside another sidebar group's dropdown, and if so, it hides the icon and only shows the heading.

Screenshot 2025-09-15 at 02 58 44PM@2x

Fixes livewire/flux#1936


slovenianGooner:
@joshhanley Great solution, perhaps for the menu to behave exactly the same as a flux:menu a chevron on the right would be appropriate in case of a submenu.

Screenshot 2025-09-15 at 07 19 09

calebporzio:
I had the same thought as @slovenianGooner - is this a reasonable feat @joshhanley ?

calebporzio:
I'm going to merge this and it can be separate addition in the next release as to not block this change.

slovenianGooner:
Another quick comment on this one - have a feeling the icon should also be available - if defined. Currently, the offset does not look OK. But otherwise, it works as expected.

Screenshot 2025-09-16 at 11 02 19

joshhanley:
@slovenianGooner I’m off for the rest of the week. Any chance you could open an issue for this so we don’t miss it? Thanks!

Issue #1943, opened 2025-09-10
Modal 'close' <button> element incorrectly uses 'alt' attribute.


FehrenbachMark:
### Flux version

2.0

### Livewire version

3.4

### Tailwind version

4.07

### Browser and Operating System

Chrome/Windows 11

### What is the problem?

The flux:modal 'close modal' <button> contains an alt attribute (alt="Close modal").

### Code snippets to replicate the problem

```blade

<button type="button" class="relative items-center font-medium justify-center gap-2 whitespace-nowrap disabled:opacity-75 dark:disabled:opacity-75 disabled:cursor-default disabled:pointer-events-none h-8 text-sm rounded-md w-8 inline-flex  bg-transparent hover:bg-zinc-800/5 dark:hover:bg-white/15 text-zinc-800 dark:text-white      text-zinc-400! hover:text-zinc-800! dark:text-zinc-500! dark:hover:text-white!" data-flux-button="data-flux-button" alt="Close modal">
        <svg class="shrink-0 [:where(&amp;)]:size-5" data-flux-icon="" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20" fill="currentColor" aria-hidden="true" data-slot="icon">
  <path d="M6.28 5.22a.75.75 0 0 0-1.06 1.06L8.94 10l-3.72 3.72a.75.75 0 1 0 1.06 1.06L10 11.06l3.72 3.72a.75.75 0 1 0 1.06-1.06L11.06 10l3.72-3.72a.75.75 0 0 0-1.06-1.06L10 8.94 6.28 5.22Z"></path>
</svg>
    </button>

```

### Screenshots/ screen recordings of the problem

<img width="350" height="176" alt="Image" src="https://github.com/user-attachments/assets/9a669e66-6a2b-40b3-880d-cf2fcb1755ed" />

### How do you expect it to work?

This should likely be an 'aria-label' instead of an alt attribute, so it will be read aloud by screen readers.

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

joshhanley:
@FehrenbachMark thanks for reporting! I've submitted a PR with a fix.

PR #1950, opened 2025-09-15
Fix modal close button aria label

joshhanley:
# The scenario

Currently the modal close button doesn't have the correct label for screen readers, it just reads "Button".

image

# The problem

The issue is that the close button has `alt="Close modal"` but it should be `aria-label="Close modal"`.

# The solution

This PR changes the `alt="Close modal"` to `aria-label="Close modal"

image

Fixes livewire/flux#1943


calebporzio:
perfect, tyty

Issue #1920, opened 2025-09-05
Sidebar.search icon position + kbd icon


Webotvorba:
### Flux version

2.3.1

### Livewire version

3.6.4

### Tailwind version

4.1.11

### Browser and Operating System

safari on macos

### What is the problem?

The search icon is misplaced and the kbd attribute is ignored. 

### Code snippets to replicate the problem

```blade
<flux:sidebar sticky collapsible class="bg-zinc-50 dark:bg-zinc-900 border-r rtl:border-r-0 rtl:border-l border-zinc-200 dark:border-zinc-700">
            <flux:sidebar.header>
                <flux:sidebar.brand href="{{ route('admin.dashboard') }}" logo="/images/miit.png" logo:dark="/images/miit.png" name="MiiT E-commerce" class="px-2" />
                <flux:sidebar.collapse class="in-data-flux-sidebar-on-desktop:not-in-data-flux-sidebar-collapsed-desktop:-mr-2" />
            </flux:sidebar.header>

            <flux:modal.trigger name="search" shortcut="cmd.k">
                <flux:sidebar.search as="button" placeholder="Search..." icon="magnifying-glass" kbd="⌘K" />
            </flux:modal.trigger>
</flux:sidebar>
```

### Screenshots/ screen recordings of the problem

<img width="327" height="158" alt="Image" src="https://github.com/user-attachments/assets/e5aff580-4c03-474e-9e33-bcad039001ce" />

### How do you expect it to work?

The input should look the same as:

```blade
<flux:input as="button" placeholder="Search..." icon="magnifying-glass" kbd="⌘K" />
```

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

joshhanley:
@Webotvorba thanks for reporting! I've just tested this and the alignment looks good to me. 

<img width="1168" height="688" alt="Image" src="https://github.com/user-attachments/assets/5d3e2141-5111-4809-8a76-9348213530fd" />

There currently is no keyboard attribute support on the sidebar search component. I'll double check to see if that was intended or not, and add it if not.

Webotvorba:
Could you test it on laravel livewire starter kit?

joshhanley:
@Webotvorba I've just tested this on Laravel's Livewire starter kit and it looks ok there to me too.

[...]

PR #1952, opened 2025-09-15
Add sidebar search `kbd` prop

joshhanley:
# The scenario

Currently if you add a keyboard shortcut to the sidebar search, nothing happens.

```blade
<flux:sidebar.search placeholder="Search..." kbd="⌘K" />
```

Screenshot 2025-09-15 at 06 36 42PM@2x

# The problem

The issue is that the new sidebar search component doesn't have a `kbd` prop, but it does have a `tooltipKbd` prop.

# The solution

This PR adds support for a `kbd` prop to display the keyboard shortcut on the search component in the expanded state, and if no `tooltipKbd` has been included, it will also set it to the passed in `kbd` value, so you don't need to set both.

Screenshot 2025-09-15 at 06 36 26PM@2x

Screenshot 2025-09-15 at 06 38 43PM@2x

Fixes livewire/flux#1920

### DIFF
--- a/stubs/resources/views/flux/sidebar/search.blade.php
+++ b/stubs/resources/views/flux/sidebar/search.blade.php
@@ -9,10 +9,13 @@
     'placeholder' => __('Search...'),
     'tooltipKbd' => null,
     'tooltip' => null,
+    'kbd' => null,
 ])
 
 @php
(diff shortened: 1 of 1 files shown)


calebporzio:
great, tyty

PR #2028, opened 2025-10-05
Remove `@blaze` directive from file input component

o-psi:
The `@pure` directive causes Blade compilation errors when used with Livewire Blaze due to the component containing dynamic translation helpers ({!! __() !!}) and PHP conditionals.

Error: 'Cannot end a section without first starting one'

This component cannot be safely optimized at compile-time because:
1. It uses translation helpers that need runtime evaluation
2. It contains PHP conditionals (if/else) for multiple file selection
3. The dynamic wire:model attribute binding requires runtime context

Removing `@pure` allows the component to render correctly while still maintaining full functionality.


joshhanley:
@o-psi we're currently working on Blaze compatibility improvements. As such we don't want to remove `@blaze` from the input if possible, so I'm going to close this for now. Can you please open an issue so the team can investigate it? Thanks!

Issue #1891, opened 2025-08-25
opening dropdowns and modals make right fixed elements jump


mauritskorse:
### Flux version

v2.2.5

### Livewire version

v3.6.4

### Tailwind version

v4.1.12

### Browser and Operating System

chrome

### What is the problem?

I have an element on the page that I want to fix to the right. However, when a popover or modal is opened, the entire element jumps further to the right (due to disabling the scrollbar). Especially when the element is rather large, this can be quite disturbing. Currently there is no easy way to account for this issue.

### Code snippets to replicate the problem

```blade
<?php

use Livewire\Volt\Component;
new class extends Component {

}; ?>
{{-- ensure there is a scrollbar on the page --}}
<div class="flex flex-col space-y-6 items-center pt-32 h-full w-full min-h-[120vh]">

    <flux:card class="fixed right-10">
        Some right fixed element
    </flux:card >

    <flux:dropdown>
        <button>trigger me</button>
        <flux:popover>
            <div>
                This is the popover content.
            </div>
        </flux:popover>
    </flux:dropdown>

    <flux:modal.trigger name="my-modal">
        <flux:button>Open Modal</flux:button>
    </flux:modal.trigger>
    <flux:modal name="my-modal">
        <div>
            This is the modal content.
        </div>
    </flux:modal>

</div>
```

### Screenshots/ screen recordings of the problem

https://github.com/user-attachments/assets/a3c80774-f355-4d49-83fa-131ee6d720ac

### How do you expect it to work?

The element should not jump further to the right when the scrollbar is disabled.

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

mauritskorse:
I couldn't really think of an easy way to fix this except in case there would be a css variable set by Flux when the scrollbar is disabled (ie `--noscroll-padding`). Currently Flux already inserts a style attribute on the html tag to disable the scrollbar and add some padding: `style="padding-right: calc(8px); overflow: hidden;"`. Adding this variable seems to be easy. Once the scrollbar becomes visible again, the variable should be set back to 0.

That would allow you to account for this disabled scrollbar behaviour with just some css: `right-[var(--noscroll-padding)]` or probably something like: `right-[calc(--spacing(4)+var(--noscroll-padding))] `.

arifbudimanar:
[...]

Issue #1639, opened 2025-05-29
Button's loading icon not showing if `wire:click` has space after the method name


jtesta66:
### Flux version

v2.1.6

### Livewire version

v3.6.3

### Tailwind version

v4.0.8

### Browser and Operating System

Chrome on Windows

### What is the problem?

When using wire:click="function (argument...)" and expecting a loading icon on the button, the space before the function arguments prevents the loading icon from rendering

This works:
`<flux:button wire:click="dosomething(argument)">Do Something</flux:button>
`
This doesn't work:
`<flux:button wire:click="dosomething (argument)">Do Something</flux:button>`

### Code snippets

It should be very easy to reproduce with examples I provided.  The actual method running in the component doesn't matter.

Component:

```
function doSomething ($whatever) {
   sleep (5);
}
```

blade template:
`<flux:button wire:click="doSomething ('testing')">Do Something</flux:button>`

### How do you expect it to work?

I would expect to be able to have a space before the method parameters and still get the loading indicator on the button.

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

ju5t:
This is not an issue with Flux but it's got to do with Livewire itself.

```blade
<button wire:click="foo (100)" wire:loading.remove wire:target="foo (100)">test</button>
```

Personally, I'm not sure if I'd call it a bug. There isn't a single popular coding style guide that recommends using spaces between the method's name and its properties. You would only use that for control structures, like `if`, `match`, `for`, etc. Even WordPress doesn't, which has the weirdest style guide I know about.

Not my decision though :)

jtesta66:
Our entire team uses a space before function parentheses.  I'm not sure whether that's popular or not is relevant.

It is a bug; the code is/should be equivalent.  I can't think of a language that would care whether there's a space there or not. 

What you're saying is that the code behind Livewire should care (for some reason? it doesn't elsewhere ...)

Also, yeah, Josh told me this was a Livewire issue, not a Flux issue, but said I could leave it here for now.

ju5t:
@jtesta66 that's why I said it's my personal opinion, that's not necessarily the same as yours or the maintainers. It's OK that we disagree. I'm not against fixing it by the way.

[...]

PR #2036, opened 2025-10-08
Update flux.css to fix error of scroll bar and loading mode of button

fdteam-deanie:
# The scenario

- Prepare a dropdown like avatar:

`<flux:select variant="listbox" label="Assign to">
    <flux:select.option selected>
        <div class="flex items-center gap-2 whitespace-nowrap">
            <flux:avatar circle size="xs" src="https://unavatar.io/github/calebporzio" /> Caleb Porzio
        </div>
    </flux:select.option>
</flux:select>`

- Prepare a submit type button to make "loading" style when clicked:

`<flux:button variant="primary" icon="paper-airplane" class="w-full" type="submit">Submit</flux:button>`

# The problem

- Scroll bar is disappeared and screen width is increased which will break the layout when click at dropdown or a select element.
- Perform a loading mode of a button does not disable text and show loading icon as it should be.

# The solution

- Override css to ensure Flux dropdowns are not clipped by page container scroll with "important" and prevent layout shift when Flux unlocks body scroll (avoid padding compensation).
- Show and center the Flux loading spinner inside disabled buttons and keep label text visible and collapse gap while loading (when button is disabled).

Fixes livewire/flux#

### DIFF
--- a/dist/flux.css
+++ b/dist/flux.css
@@ -240,3 +240,33 @@ Utilties
 .flux-no-scrollbar::-webkit-scrollbar{
     display:none
 }
+
+/* Show and center the Flux loading spinner inside disabled buttons */
+button[disabled] [data-flux-loading-indicator] {
+    position: relative !important;
+    inset: auto !important;
+    opacity: 1 !important;
+    display: inline-flex !important;
+    align-items: center !important;
+    justify-content: center !important;
+    vertical-align: middle;
+    transform: translateY(1px);
+}
+
+/* Keep label text visible and collapse gap while loading (when button is disabled) */
+button[disabled] > span {
+    opacity: 1 !important;
+    gap: 0 !important;
+}
+
+/* Ensure Flux dropdowns are not clipped by page container scroll */
+html,
+body,
+[data-flux-page-container] {
+    overflow: visible !important;
+}
+
+/* Prevent layout shift when Flux unlocks body scroll (avoid padding compensation) */
+html[data-flux-scroll-unlock] {
+    padding-right: 0 !important;
+}


joshhanley:
@fdteam-deanie thanks for the PR! Can you please provide a screen recording of the issue/ before the fix and one after, so we can see the difference? Add the before into the scenario section and after into the solution section. Thanks!

joshhanley:
@fdteam-deanie also this seems like two separate problems you are trying to fix. Can you please split them into two separate PRs with details and screen recordings about each problem? Thanks!

fdteam-deanie:
> @fdteam-deanie thanks for the PR! Can you please provide a screen recording of the issue/ before the fix and one after, so we can see the difference? Add the before into the scenario section and after into the solution section. Thanks!

Here you go:

- Before this PR (as https://github.com/livewire/flux/issues/1891 and https://github.com/livewire/flux/issues/1639):

[https://www.youtube.com/watch?v=atgp7H8RcDo](https://www.youtube.com/watch?v=atgp7H8RcDo)

- After this PR:

[https://www.youtube.com/watch?v=39EEp88NogE](https://www.youtube.com/watch?v=39EEp88NogE)

joshhanley:
@fdteam-deanie please upload the videos to the PR/ GitHub as we don't want to open random Google drive links. Thanks!

fdteam-deanie:
> @fdteam-deanie also this seems like two separate problems you are trying to fix. Can you please split them into two separate PRs with details and screen recordings about each problem? Thanks!

I think so, they should be 2 different problems. But all come from the problem of Tailwind css.

fdteam-deanie:
> @fdteam-deanie please upload the videos to the PR/ GitHub as we don't want to open random Google drive links. Thanks!

I tried but failed to upload recorded screen videos to this. I uploaded to Youtube, safe for you:

- Before this PR (as https://github.com/livewire/flux/issues/1891 and https://github.com/livewire/flux/issues/1639):

[https://www.youtube.com/watch?v=atgp7H8RcDo](https://www.youtube.com/watch?v=atgp7H8RcDo)

- After this PR:

[https://www.youtube.com/watch?v=39EEp88NogE](https://www.youtube.com/watch?v=39EEp88NogE)

ju5t:
> I think so, they should be 2 different problems. But all come from the problem of Tailwind css.

@fdteam-deanie it's about what the changes do, not what file you made the changes in. It's easier for maintainers if a PR is limited in scope, for example, the loading indicator on the button.

But it's also better for the people who raise PRs. For example, if the maintainers don't want to merge your entire PR they might reject it altogether. But if you would have raised them as individual PRs (one per problem), it's likely they'd merge the change they like and just reject the ones they don't.

PR #2037, opened 2025-10-09
Update flux.css to fix error of dropdown

fdteam-deanie:
Fix error that the scroll bar of the page is disappeared and screen width is increased which will break the page layout when click at a dropdown element or a `<select>` element.

# The scenario

1. Prepare a dropdown like avatar:

`<flux:select variant="listbox" label="Assign to"> <flux:select.option selected> <div class="flex items-center gap-2 whitespace-nowrap"> <flux:avatar circle size="xs" src="https://unavatar.io/github/calebporzio" /> Caleb Porzio </div> </flux:select.option> </flux:select>`

2. Make sure your page has content that need to scroll (or you can zoom it in until scroll bar appears).

3. Click the dropdown or any selectable element, then you see the scroll bar of the page (usually on the right side) disappear and all your page's elements "jump" to the right.

# The problem

- Scroll bar is disappeared and screen width is increased which will break the layout when click at dropdown or a select element.
- Reported at https://github.com/livewire/flux/issues/1891.
- Video of the error before apply PR:

https://github.com/user-attachments/assets/5ea0b027-c148-4e04-a061-e8a9fcc50ade

# The solution

- Override css to ensure Flux dropdowns are not clipped by page container scroll with "important" and prevent layout shift when Flux unlocks body scroll (avoid padding compensation).
- Video of successfully fixed after apply PR:

https://github.com/user-attachments/assets/8982befc-047c-4759-b4f3-cb375b16df8e

Fixes livewire/flux#

### DIFF
--- a/dist/flux.css
+++ b/dist/flux.css
@@ -240,3 +240,15 @@ Utilties
 .flux-no-scrollbar::-webkit-scrollbar{
     display:none
 }
+
+/* Ensure Flux dropdowns are not clipped by page container scroll */
+html,
+body,
+[data-flux-page-container] {
+    overflow: visible !important;
+}
+
+/* Prevent layout shift when Flux unlocks body scroll (avoid padding compensation) */
+html[data-flux-scroll-unlock] {
+    padding-right: 0 !important;
+}


joshhanley:
@fdteam-deanie thanks for the PR! I just tested the issue reported in #1891 with your fix, and while it stops the fixed element from moving around, it has now re-enabled scroll behind modals, which is what we had blocked. See below screen capture. Because it doesn't fix the issue while maintaining scroll lock functionality, I'm going to close this. Thanks for being willing to contribute! Please feel free to open a new PR if you come up with any other ideas on how we can fix it.

Screenshot 2025-11-25 at 02 02 36PM

PR #2043, opened 2025-10-11
Deduplicate package check in `ensurePro` method

sajjadhossainshohag:
# The scenario

When reviewing the `FluxManager` class, I noticed that the `ensurePro()` method directly calls `InstalledVersions::isInstalled('livewire/flux-pro')` to check if Flux Pro is installed, while there's already a dedicated `pro()` method that performs the same check.

# The problem

deduplicate package check in `ensurePro` method

# The solution

Refactor `ensurePro()` to call the existing `pro()` method instead of duplicating the package check

Fixes livewire/flux#

### DIFF
--- a/src/FluxManager.php
+++ b/src/FluxManager.php
@@ -26,7 +26,7 @@ class FluxManager
 
     public function ensurePro()
     {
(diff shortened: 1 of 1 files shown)


joshhanley:
Thanks!

Issue #2031, opened 2025-10-07
Avatar overflow and cover


cincer:
### Flux version

v2.5.0

### Livewire version

v3.6.4

### Tailwind version

v4.1.14

### Browser and Operating System

Safari on macOS

### What is the problem?

When I upload not squared image for avatar it displayed wrong with flux:sidebar.profile and flux:avatar also. 
If I add these classes it works well, but it should be included automatically in the component:  

class="[&_img]:size-full [&_img]:object-cover"

### Code snippets to replicate the problem

<flux:avatar size="xl"
                    class="[&_img]:size-full [&_img]:object-cover"
                    :name="$row->name"
                    :src="$row->src" />

### Screenshots/ screen recordings of the problem

<img width="417" height="180" alt="Image" src="https://github.com/user-attachments/assets/3f699385-29e5-4113-912d-3e634bd39e0f" />

### How do you expect it to work?

<img width="406" height="182" alt="Image" src="https://github.com/user-attachments/assets/f40ae32c-86be-4b32-88fd-08b2a5349606" />

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

joshhanley:
@cincer thanks for reporting! I've submitted a PR with a fix.

PR #2047, opened 2025-10-13
Fix avatar image cover

joshhanley:
# The scenario

Currently if you have an avatar or profile component and the provided image isn't perfectly square, the displayed image doesn't look right and the background colour is being shown.

Screenshot 2025-10-13 at 10 40 53AM@2x

```blade
<?php

use Livewire\Component;

new class extends Component {
    //
};

?>

<div class="max-w-192 mx-auto p-6 space-y-4">
    <flux:avatar src="https://plus.unsplash.com/premium_photo-1760018690381-a8a38d50b55a?ixlib=rb-4.1.0&ixid=M3wxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8fA%3D%3D&auto=format&fit=crop&q=80&w=3271" />
    <flux:profile avatar="https://plus.unsplash.com/premium_photo-1760018690381-a8a38d50b55a?ixlib=rb-4.1.0&ixid=M3wxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8fA%3D%3D&auto=format&fit=crop&q=80&w=3271" />
</div>
```

# The problem

The issue is that the image is being shrunk to match the size of the wrapping element. But as it's not square, this means that the background is being shown above/below the image.

# The solution

The solution is to add `size-full object-cover` to the `` element which ensures it fits the available space.

Screenshot 2025-10-13 at 10 38 56AM@2x

Fixes livewire/flux#2031

### DIFF
--- a/stubs/resources/views/flux/avatar/index.blade.php
+++ b/stubs/resources/views/flux/avatar/index.blade.php
@@ -173,7 +173,7 @@ $label = $alt ?? $name;
 <flux:with-tooltip :$tooltip :$attributes>
     <flux:button-or-link :attributes="$attributes->class($classes)->merge($circle ? ['data-circle' => 'true'] : [])" :$as :$href data-flux-avatar data-slot="avatar" data-size="{{ $size }}">
         <?php if ($src): ?>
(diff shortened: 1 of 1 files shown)


calebporzio:
simple. good. yep.

Issue #2021, opened 2025-10-02
Tooltip causing layout issue inside button group with modal trigger


DucPanigale:
### Flux version

v2.4.0

### Livewire version

v3.6.4

### Tailwind version

v4.0.7

### Browser and Operating System

Edge on Windows 11

### What is the problem?

Hi y'all! 

First of all, I love Flux - it's such a sleek, clean layout. It's fun to work with it.
I have a tiny issue:
If I put a button with a tooltip inside a modal trigger, the button renders as a separate button like in my screenshot the top button group. Is this possibly related to issue #161 ? 
Thank you very much in advance.

Kind regards

### Code snippets to replicate the problem

```blade
<flux:button.group>
    <flux:button size="sm" icon="pencil"> </flux:button>
     <flux:modal.trigger name="page-details-{{ $page->id }}">
        <flux:tooltip content="{{ __('Trash') }}" position="bottom">
            <flux:button size="sm" icon="trash">
            </flux:button>
        </flux:tooltip>
    </flux:modal.trigger>
</flux:button.group>
<div class="mb-5">
</div>
<flux:button.group>
    <flux:button size="sm" icon="pencil"> </flux:button>
     <flux:modal.trigger name="page-details-{{ $page->id }}">
        <flux:button size="sm" icon="trash">
        </flux:button>
    </flux:modal.trigger>
</flux:button.group>
```

### Screenshots/ screen recordings of the problem

<img width="80" height="110" alt="Image" src="https://github.com/user-attachments/assets/ee9cf011-beb1-48e2-b580-9a72ea74943f" />

### How do you expect it to work?

I would expect it to not change the button borders and basically look like the second button group in the screenshot.

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

joshhanley:
@DucPanigale thanks for reporting! I've submitted a PR with a fix.

joshhanley:
@DucPanigale the original PR didn't end up working, so I've submitted a new PR #2404 with a fix for this.

PR #2048, opened 2025-10-13
Fix button group with modal trigger

joshhanley:
# The scenario

Currently if you have a button group, with buttons that have both a tooltip and are a modal trigger, then the buttons borders aren't rendered correctly to match the group.

Screenshot 2025-10-13 at 02 12 33PM@2x

```blade
<flux:button.group>
    <flux:modal.trigger name="page-details">
        <flux:button icon="bars-3-bottom-left" tooltip="Left align" />
    </flux:modal.trigger>
    <flux:modal.trigger name="page-details">
        <flux:button icon="bars-3" tooltip="Center align" />
    </flux:modal.trigger>
    <flux:modal.trigger name="page-details">
        <flux:button icon="bars-3-bottom-right" tooltip="Right align" />
    </flux:modal.trigger>
</flux:button.group>
```

# The problem

The issue is that the CSS selectors which remove the rounded and excess borders on inner buttons is scoped to only apply these styles for immediate children of the button group. There are also selectors which account for a button group with a tooltip that contains a button (1 level of nesting).

But when using a modal trigger, the button itself ends up a few levels below the button group in the DOM and as such isn't being targeted by the child selectors.

# The solution

The solution is to update the selectors to detect any buttons within the group and apply the correct styles.

So for example, this PR changes the selector from:

```
[&>*:last-child:not(:first-child)>[data-flux-group-target]:not([data-invalid])]:border-s-0
```

To

```
[&>*:last-child:not(:first-child)_[data-flux-group-target]:not([data-invalid])]:border-s-0
```

The main difference is before the `[data-flux-group-target]` instead of a `>` which is direct children, we now use `_` which means any children.

One bonus we get by making this change is we had a selector for `sub-sub-children of input groups` which is now no longer needed as it is covered by these selector changes.

Now it all works as expected.

Screenshot 2025-10-13 at 02 12 07PM@2x

I also added this scenario to our `groups` test page and checked all the other groups still worked as expected.

Screenshot 2025-10-13 at 02 11 31PM@2x

Fixes livewire/flux#2021
Fixes livewire/flux#2072

### DIFF
--- a/stubs/resources/views/flux/button/group.blade.php
+++ b/stubs/resources/views/flux/button/group.blade.php
@@ -15,9 +15,9 @@ $classes = Flux::classes()
         '[&>[data-flux-input]:has(+[data-flux-input-group-suffix])>[data-flux-group-target]:not([data-invalid])]:border-e-0',
 
         // Selects and date pickers borders...
-        '[&>*:last-child:not(:first-child)>[data-flux-group-target]:not([data-invalid])]:border-s-0',
-        '[&>*:not(:first-child):not(:last-child)>[data-flux-group-target]:not([data-invalid])]:border-s-0',
-        '[&>*:has(+[data-flux-input-group-suffix])>[data-flux-group-target]:not([data-invalid])]:border-e-0',
+        '[&>*:last-child:not(:first-child)_[data-flux-group-target]:not([data-invalid])]:border-s-0',
+        '[&>*:not(:first-child):not(:last-child)_[data-flux-group-target]:not([data-invalid])]:border-s-0',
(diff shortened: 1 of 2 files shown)


calebporzio:
Oooh hmm. I think I avoided descendant selectors because people might have like a menu inside the input group or something else like a popover and now something nested inside of it might be impacted?

Want to just sanity check me on that and make sure there won't be scenarios where these styles will impact things they shouldn't? 

If not, let's ship it. good improvement!

joshhanley:
Yeah, sadly it does impact button inside menus 😕 will have to think of another solution.

joshhanley:
Closing in favour of PR #2404

Issue #1998, opened 2025-09-29
Styling of file input within input group


mauritskorse:
### Flux version

2.4.0

### Livewire version

3.6.4

### Tailwind version

4.1.13

### Browser and Operating System

Chrome on Windows

### What is the problem?

When having an file input within the `input.group` the styling of the file name part is not affected. Even though the button for selecting the file is styled accordingly.

Current workaround is by adding the following classes to the input group:
```blade
<flux:field>
    <flux:input.group class="[&>[data-flux-input-file]]:gap-0! [&>[data-flux-input-file]>[x-ref=name]]:h-10 [&>[data-flux-input-file]>[x-ref=name]]:border-t [&>[data-flux-input-file]>[x-ref=name]]:border-b [&>[data-flux-input-file]>[x-ref=name]]:border-zinc-200
        [&>[data-flux-input-file]>[x-ref=name]]:shadow-xs  [&>[data-flux-input-file]>[x-ref=name]]:border-s-0 [&>[data-flux-input-file]>[x-ref=name]]:ps-4 [&>[data-flux-input-file]>[x-ref=name]]:pe-4 [&>[data-flux-input-file]>[x-ref=name]]:leading-9 [&>[data-flux-input-file]>[x-ref=name]:not(last-child)]:border-e
        [&>[data-flux-input-file]]:w-auto
    ">
        <flux:input type="file" wire:model="fileImport" />
        <flux:button wire:click="importFile" icon="arrow-up-tray">Import</flux:button>
    </flux:input.group>
</flux:field>
```

### Code snippets to replicate the problem

```blade
<flux:field>
    <flux:input.group>
        <flux:input type="file" wire:model="fileImport" />
        <flux:button wire:click="importFile">Import</flux:button>
    </flux:input.group>
</flux:field>
```

### Screenshots/ screen recordings of the problem

<img width="348" height="53" alt="Image" src="https://github.com/user-attachments/assets/4010b1f1-309c-4af3-8585-ff87c29c40ca" />

### How do you expect it to work?

I expect it to look somewhat like this:

<img width="345" height="52" alt="Image" src="https://github.com/user-attachments/assets/fc4d4cc6-c48a-4086-b3c7-8b0e24583370" />

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

[...]

PR #2049, opened 2025-10-13
Fix file input in a group

joshhanley:
# The scenario

Currently, if you use a basic file input in an input group, the file input does not match the style of the group properly and the group on the right looks broken.

Screenshot 2025-10-13 at 03 46 19PM@2x

```blade
<?php

use Livewire\Component;

new class extends Component {
    //
};

?>

<div class="max-w-192 mx-auto p-6 space-y-4">
    <flux:field>
        <flux:label>File input in a group</flux:label>
        <flux:input.group>
            <flux:input type="file" wire:model="fileImport" />
            <flux:button icon="arrow-up-tray" wire:click="importFile">Import</flux:button>
        </flux:input.group>
    </flux:field>
</div>
```

# The problem

The issue is that the file name part of the file input doesn't have any styles applied to it that would make it look like an input or button (no border, shadow, etc).

So when it gets included in an input group, it then just looks out of place.

# The solution

The solution is to add custom styles to the file name part of the file input so it is styled correctly when used in a group. These should only be applied when it's in a group and not when used standalone.

**Light mode**
Screenshot 2025-10-13 at 03 51 12PM@2x

I also made sure the styles worked in dark mode. But one thing that didn't look great is the contract between the file name and the background colour.

**Dark mode with existing text colour in file name section**
Screenshot 2025-10-13 at 03 52 07PM@2x

So I made the file name text one shade lighter when it is used in a group, so the contrast is better between it and the background.

**Dark mode with slightly lighter text colour in file name section**
Screenshot 2025-10-13 at 03 51 37PM@2x

Fixes livewire/flux#1998

### DIFF
--- a/stubs/resources/views/flux/input/file.blade.php
+++ b/stubs/resources/views/flux/input/file.blade.php
@@ -17,7 +17,7 @@ extract(Flux::forwardedAttributes($attributes, [
 @php
 $classes = Flux::classes()
     ->add('w-full flex items-center gap-4')
-    ->add('[[data-flux-input-group]_&]:w-auto [[data-flux-input-group]_&]:items-stretch [[data-flux-input-group]_&]:gap-0')
+    ->add('[[data-flux-input-group]_&]:items-stretch [[data-flux-input-group]_&]:gap-0')
 
     // NOTE: We need to add relative positioning here to prevent odd overflow behaviors because of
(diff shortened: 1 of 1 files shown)


calebporzio:
Wow, nice work. I don't love all those in-group prefixes in the div, but i don't see a better way. probably best to go with the literal thing like that for now. if this was a more versatile component I might see if we can find something better, but because it's kinda one-off and well scoped i saw we do it.

PR #2051, opened 2025-10-13
Make the new pro components publishable

dougffer:
# The scenario

I'm trying to fix issues in the file-upload and file-item components. To do this, I need to make some changes.

I also noticed that Pillbox was missing from the list, and that Callout is in the Pro group instead of Free.

# The problem

File upload, file item and pillbox are not publishable.

# The solution

Add them to the list of components.


joshhanley:
@dougffer thanks!

PR #2078, opened 2025-10-23
Fix sidebar search icon alignment

ganyicz:
# The problem

The search icon in the sidebar is not aligned in the center.

SCR-20251023-sygk

Before / After

# The solution

Added `items-center` to the button.

Additionally, the PR removes conflicting classes:

SCR-20251023-taww


calebporzio:
Thanks! Good catch

PR #2088, opened 2025-10-29
Add attributes to modal close component

olyviero:
# The scenario

I found impossible to have a button in full width inside the <flux:modal.close> component.

# The problem

It is impossible to pass classes to the <flux:modal.close> component.

# The solution

I added the class prop.

Fixes livewire/flux#

### DIFF
--- a/stubs/resources/views/flux/modal/close.blade.php
+++ b/stubs/resources/views/flux/modal/close.blade.php
@@ -1,5 +1,3 @@
-@props(['class' => ''])
-
 @blaze
 
 <ui-close data-flux-modal-close {{ $attributes }}>


stubs/resources/views/flux/modal/close.blade.php
```diff
@@ -1,5 +1,7 @@
+@props(['class' => ''])
+
 @blaze
 
-<ui-close data-flux-modal-close>
+<ui-close data-flux-modal-close {{ $attributes->merge(['class' => $class]) }}>
```
joshhanley: This can just be `{{ $attributes }}` as class will be forwarded automatically.

```suggestion
<ui-close data-flux-modal-close {{ $attributes }}>
```

stubs/resources/views/flux/modal/close.blade.php
```diff
@@ -1,5 +1,7 @@
+@props(['class' => ''])
+
```
joshhanley: A class prop isn't needed. See below.

```suggestion
```

stubs/resources/views/flux/modal/close.blade.php
olyviero: @joshhanley  Absolutely cleaner ! Thanks :)
Users can implement defaults as they need it...

calebporzio:
Thanks!

Issue #2056, opened 2025-10-15
"field:class" scoped attribute stopped working on 2.6.0


squaredpx:
### Flux version

v2.6.0

### Livewire version

v3.6.4

### Tailwind version

v4.1.14

### Browser and Operating System

Chrome on MacOS

### What is the problem?

Up until v2.5.1 scoped attributes with `field:class` were enabled as a way to pass classes to a component that are applied to the wrapping field component instead of being applied to the input when using the short hand syntax (see #1429)

v2.6.0 broke this and now those classes are not being applied to the parent. I think the issue comes from [this commit](https://github.com/livewire/flux/commit/a4d670f6b3545111c79c5378e10a22694530dbca)

### Code snippets to replicate the problem

```
<?php

use function Livewire\Volt\{state};

//

?>

<div class="grid grid-cols-4 gap-2 border border-red-500 p-2">
    <flux:input label="Input 1" field:class="col-span-2"/>
    <flux:input label="Input 2" field:class="col-span-2"/>
</div>
```

### Screenshots/ screen recordings of the problem

In the example provided a `col-span-2` is applied to the flux input components via `field:class` so they span 2 columns each but it's not passed to the parent wrapper so it's not applied

![Image](https://github.com/user-attachments/assets/ec2b18b5-2c79-45be-9af9-8fb315c29197)

### How do you expect it to work?

The classes passed using `field:class` should be applied to the parent so in the example both fields span 2 columns

![Image](https://github.com/user-attachments/assets/76da9a22-be71-4fce-bedf-4d642d70d4b4)

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

joshhanley:
@squaredpx thanks for reporting! I've submitted a PR with a fix.

PR #2089, opened 2025-10-29
Fix `Flux::attributesAfter()` usages

joshhanley:
# The scenario

Currently if you have a `label:class` attribute on an input, like `<flux:input label="Input 1" label:class="text-purple-500"/>`, the classes are no longer forwarded since v2.6.0, where as it used to work.

Screenshot 2025-10-29 at 07 21 50PM@2x

```blade
<?php

use Livewire\Component;

new class extends Component {
    //
};

?>

<div class="mx-auto max-w-md">
    <flux:input label="Input 1" label:class="text-purple-500"/>
</div>
```

# The problem

The issue is that in v2.6.0 we released an update to `Flux::attributesAfter()` that stripped the matched values from existing `$attributes` to ensure they weren't also output/ doubled up where they shouldn't be.

The problem with stripping comes down to how `Flux::attributesAfter()` is used. One scenario works fine and the other is causing the problem described above.

The working solution is when used like this:

```blade
<?php $labelAttributes = Flux::attributesAfter('label', $attributes); ?>
<flux:label :attributes="$labelAttributes" />
```

The broken scenario is when it is used inline like this:

```blade
<flux:label :attributes="Flux::attributesAfter('label', $attributes)" />
```

To understand why this is broken, we have to dive into how Blade compiles components.

When a Blade component is found, Blade compiles it down to this string:

```php
<?php $component = Illuminate\View\AnonymousComponent::resolve(['view' => 'ab18b3e58a3b1bb5106ced208a8bd460::label','data' => ['attributes' => Flux::attributesAfter('label:', $attributes, ['badge' => $badge])]] + (isset($attributes) && $attributes instanceof Illuminate\View\ComponentAttributeBag ? $attributes->all() : [])); ?>
<?php $component->withName('flux::label'); ?>
<?php if ($component->shouldRender()): ?>
<?php $__env->startComponent($component->resolveView(), $component->data()); ?>
<?php if (isset($attributes) && $attributes instanceof Illuminate\View\ComponentAttributeBag): ?>
<?php $attributes = $attributes->except(\Illuminate\View\AnonymousComponent::ignoredParameterNames()); ?>
<?php endif; ?>
<?php $component->withAttributes(['attributes' => \Illuminate\View\Compilers\BladeCompiler::sanitizeComponentAttribute(Flux::attributesAfter('label:', $attributes, ['badge' => $badge]))]); ?>
```

If you look closely, you will see that `Flux::attributesAfter()` is actually being called twice for the label component. This is because Blade makes use of the `$attributes` bag twice when rendering a component.

What this means is that the first call is removing the label attributes from the `$attribute` bag. And the second one inside `->withAttributes()`, which is the one that actually passes the data to the component, can't actually find any label attributes in the `$attributes` bag. Hence why no attributes are being forwarded to the component.

# The solution

The solution is to remove the inline usages of `Flux::attributesAfter()` and make use of a temporary variable, to ensure the attributes are pulled and collected only once and then that bag is included with the component.

I have checked Flux Pro and all components that use `Flux::attributesAfter()` are already using temporary variables, so there was only a couple of components that needed to be updated on this repo.

Now it's all working as expected again.

Screenshot 2025-10-29 at 07 35 03PM@2x

Fixes livewire/flux#2056

### DIFF
--- a/stubs/resources/views/flux/navbar/item.blade.php
+++ b/stubs/resources/views/flux/navbar/item.blade.php
@@ -76,6 +76,7 @@ $classes = Flux::classes()
     <?php endif; ?>
 
     <?php if (isset($badge) && $badge !== ''): ?>
-        <flux:navbar.badge :attributes="Flux::attributesAfter('badge:', $attributes, ['color' => $badgeColor, 'class' => 'ms-2'])">{{ $badge }}</flux:navbar.badge>
+        <?php $badgeAttributes = Flux::attributesAfter('badge:', $attributes, ['color' => $badgeColor, 'class' => 'ms-2']); ?>
+        <flux:navbar.badge :attributes="$badgeAttributes">{{ $badge }}</flux:navbar.badge>
     <?php endif; ?>
 </flux:button-or-link>
--- a/stubs/resources/views/flux/navlist/item.blade.php
+++ b/stubs/resources/views/flux/navlist/item.blade.php
@@ -84,6 +84,7 @@ $classes = Flux::classes()
     <?php endif; ?>
 
     <?php if (isset($badge) && $badge !== ''): ?>
-        <flux:navlist.badge :attributes="Flux::attributesAfter('badge:', $attributes, ['color' => $badgeColor])">{{ $badge }}</flux:navlist.badge>
+        <?php $badgeAttributes = Flux::attributesAfter('badge:', $attributes, ['color' => $badgeColor]); ?>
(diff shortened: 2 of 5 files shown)


sertxudev:
Can confirm this also happens with the profile component:

```blade
<flux:profile
    class="hidden md:flex"
    name="Sergio Peris"
    icon-trailing="chevron-down"
    avatar:badge avatar:badge:color="pink" avatar:badge:position="top right"
/>
```

joshhanley:
@sertxudev thanks! Yeah I've added fix profile in this PR too 🙂

PR #2092, opened 2025-10-29
Hide key visibility in `flux:activate` command

Awesomolocity:
# The scenario

You are running a conference (named "Wire⚡️Live", perhaps). Members of the audience point out that the latest version of Livewire and Flux do not seem to work well together when using the Livewire Starter Kit. 

# The problem

As you replicate the scenario, there is no easy way to hide your key from sight when entering it into the `flux:activate` command prompts.

public-example

# The solution

~~We add an optional `--private` flag. When this flag is present, the input fields for both `email` and `key` are shown as password fields, rather than text fields, which hides the inputs from audience view.~~

We change the key field from a text input to a password input, masking the entry to any onlookers.

```php
$email = ($private) ? password(...$email_args) : text(...$email_args);
//...
$key = ($private) ? password(...$key_args) : text(...$key_args);
```

Screenshot 2025-10-29 at 14 28 41

### DIFF
--- a/stubs/resources/views/flux/error.blade.php
+++ b/stubs/resources/views/flux/error.blade.php
@@ -1,8 +1,9 @@
 @props([
-    'name' => null,
+    'icon' => 'exclamation-triangle',
+    'bag' => 'default',
     'message' => null,
     'nested' => true,
-    'bag' => 'default',
+    'name' => null,
 ])
 
 @php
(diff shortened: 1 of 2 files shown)


joshhanley:
@Awesomolocity thanks for the PR! I actually think it should just be private by default. But only for the licence key. Basically the same as a login form. What do you think?

Awesomolocity:
@joshhanley Mmmm...the reason I reached for a flag was that while working on this is that I was sitting with someone who had recorded a Youtube video where he activated flux, and was talking about the woes of having to edit both the key AND the email out of the video.

However, that is admittedly a niche use-case and maybe not worth having a flag for, when changing the key to a password (and no flag at all) is the solution with the least amount of code changes.

I can go ahead and make that change, unless you have any other thoughts to add?

joshhanley (changes requested):
@Awesomolocity ah right fair! I just tested what composer does and it shows the email address and hides the password, so I think we should just do the same. Yep, if you can change it, that'd be great 🙂

Screenshot 2025-10-30 at 02 00 48PM@2x

Awesomolocity:
@joshhanley Aye aye, captain! Change has been made, should be all ready for ya. :)

calebporzio:
Thanks!

Issue #2050, opened 2025-10-13
Switch indicator not switching on specific dark/light mode config


mauritskorse:
### Flux version

v2.5.1

### Livewire version

v3.6.4

### Tailwind version

v4.1.12

### Browser and Operating System

chrome on windows

### What is the problem?

To allow certain parts of my page show either in light or dark mode while the rest of the page is in opposite mode, I have the following tailwind custom variant in my app.css. 

```css
@custom-variant dark (&:where(.dark, .dark *):not(.light, .light *,.not-dark, .not-dark *)) /* dark mode */;
```

However this breaks the toggle switch because it causes utility classes with a `dark` (or light) prefix to take precedence over non dark-mode prefixed classes. Since the switch indicator has an off-state location defined for dark mode, but no dark mode for the on-state, it will always stay at the off state location.

These are the current relevant classes for the swith element indicator:
```php
$indicatorClasses = Flux::classes()
   //...
   ->add('transition translate-x-[3px] dark:translate-x-[2px] rtl:-translate-x-[3px] dark:rtl:-translate-x-[2px]'),
   //...
   ->add([
       'group-data-checked:translate-x-[15px] rtl:group-data-checked:-translate-x-[15px]',
       'group-data-checked:bg-(--color-accent-foreground)'
   ]);
```

To fix it I have to add the following classes to each of my switches:
```blade
<flux:switch class="
    dark:data-checked:[&_span]:translate-x-[15px] 
    dark:data-checked:rtl:[&_span]:translate-x-[15px]
">
```

### Code snippets to replicate the problem

In app.css:
```css
@custom-variant dark (&:where(.dark, .dark *):not(.light, .light *,.not-dark, .not-dark *)) /* dark mode */;
```

Blade
```blade
<div class="dark p-3">
   <flux:switch label="Dark mode switch" />
</div>

<div class="light p-3">
   <flux:switch label="Lightmode switch" />
</div>
```

### Screenshots/ screen recordings of the problem

As you can see: nothing seems to be happening on the dark mode variant.

https://github.com/user-attachments/assets/38b31fb1-b688-4821-a130-bad363b9d38b

### How do you expect it to work?

I expect the dark mode to work when the dark mode variant is a little bit more advanced.
To make it work the following should be added to the `$indicatorClasses`:

`dark:data-checked:translate-x-[15px] dark:data-checked:rtl:-translate-x-[15px]`

### Please confirm (incomplete submissions will not be addressed)

[...]

PR #2146, opened 2025-11-21
Fix forced dark mode switch

joshhanley:
# The scenario

Currently if you have a forced dark/light mode, which allows you to add `dark` or `light` class on components, to just change that section of a page, the switch doesn't work in dark mode.

Screenshot 2025-11-21 at 06 40 44PM

```blade
<div>
    <div class="dark p-3 bg-zinc-900">
       <flux:switch label="Dark mode switch" />
    </div>

    <div class="light p-3">
       <flux:switch label="Lightmode switch" />
    </div>
</div>
```

# The problem

The issue is that because there is already a `dark:translate-x-[2px]` on the element, when using a dark class on an element wrapping the switch (instead of the root `<html>` element), that dark translate class takes priority over the `group-data-checked:translate-x-[15px]` which makes the switch move when toggled.

# The solution

The solution is to also add `dark:` classes for the checked translation.

```
'dark:group-data-checked:translate-x-[15px] dark:rtl:group-data-checked:-translate-x-[15px]',
```

Now everything works as expected.

Screenshot 2025-11-21 at 06 45 25PM

Fixes livewire/flux#2050

### DIFF
--- a/stubs/resources/views/flux/switch.blade.php
+++ b/stubs/resources/views/flux/switch.blade.php
@@ -30,6 +30,7 @@ $indicatorClasses = Flux::classes()
     ->add('bg-white')
     ->add([
         'group-data-checked:translate-x-[15px] rtl:group-data-checked:-translate-x-[15px]',
+        'dark:group-data-checked:translate-x-[15px] dark:rtl:group-data-checked:-translate-x-[15px]',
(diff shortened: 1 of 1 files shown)


calebporzio:
Can you add a comment above that line and just commit it to main

joshhanley:
Done!

PR #2161, opened 2025-11-24
Fix flash file-upload component on wire:navigate

arifbudimanar:
# The scenario

When using file-upload component on the page and then do wire:navigate it will flash when on dark mode.

# The problem

https://github.com/user-attachments/assets/23702c0e-998f-4295-9377-25442f5360b6

# The solution
```
ui-file-upload {
    & > * {
        transition: none;
    }
}
```

https://github.com/user-attachments/assets/549f64dc-5610-48df-9afd-de0eb2d5a80b

### DIFF
--- a/dist/flux.css
+++ b/dist/flux.css
@@ -249,7 +249,7 @@ File upload styles...
 */
 ui-file-upload {
     & > * {
-        transition: opacity 0.35s allow-discrete, transform 0.35s allow-discrete, height 0.35s allow-discrete;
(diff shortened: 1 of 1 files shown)


joshhanley:
@arifbudimanar thanks for the PR! I think this would be better fixed inside the File upload component (inside Flux Pro), rather than in the flux.css file. As such I'm going to close this PR. Can you please open an issue so the team can investigate? Thanks!

Issue #1784, opened 2025-07-11
Possible bug in field variant inline


mangrovestudios:
### Flux version

2.2

### Livewire version

3.5

### Tailwind version

4.1.11

### Browser and Operating System

Chrome on MacOS

### What is the problem?

A standard field doesn't display as inline. The classes are appearing. In fact, I changed the default in field.blade.php and still no side-by-side display.

### Code snippets to replicate the problem

```blade
<flux:field variant="inline">
  <flux:label>Email</flux:label>
  <flux:input wire:model="email" type="email" />
  <flux:error name="email" />
</flux:field>
```

### Screenshots/ screen recordings of the problem

<img width="1440" height="337" alt="Image" src="https://github.com/user-attachments/assets/f69c6942-f129-4d38-a0d0-012a91b5a8e4" />

### How do you expect it to work?

Label inline with field

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

joshhanley:
@mangrovestudios thanks for reporting! I've investigated this and the reason the inline styles are not being applied for inputs is because inputs, while they have the `data-flux-control` attribute on them, they also have a `<div>` wrapping around them which stops the selectors from working properly.

I think the `data-flux-control` attribute should be moved to the wrapping div, but Laravel's starter kit is targeting inputs using `input[data-flux-control]` so it would be a breaking change. We'll have to discuss how else we can add support for this.

The other thing I found is that by default the inline flag will apply a `grid-cols-[1fr_auto]` which works great for things like radios, checkboxes, and switches. But that doesn't work nicely for inputs and causes the label to grow and the input to shrink.

<img width="725" height="487" alt="Image" src="https://github.com/user-attachments/assets/85092af7-466d-4b3f-88f7-068d4844633f" />

So we will need to ensure that for inline inputs, that it is reversed, so that the label shrinks and the input grows to fit the available space.

<img width="725" height="487" alt="Image" src="https://github.com/user-attachments/assets/b257e53d-542f-4bfd-af8d-0f4b1cb95b6f" />

To get my above experiment working, I added the following classes to the field component:

```php
'has-[[data-flux-label]~[data-flux-input]]:grid-cols-[auto_1fr]',
'has-[[data-flux-input]~[data-flux-label]]:grid-cols-[1fr_auto]',
[...]

PR #2167, opened 2025-11-25
Add support for inline field with inputs

joshhanley:
# The scenario

Currently if you wrap an input in field and add `variant="inline"` nothing happens.

Screenshot 2025-11-25 at 01 41 03PM@2x

```blade
<flux:field variant="inline">
    <flux:label>Email</flux:label>
    <flux:input wire:model="email" type="email" />
    <flux:error name="email" />
</flux:field>
```

# The problem

When investigating this, I looked up to see how inline was being handled for radios/checkboxes and found that the following classes were added to the field component to support it.

```php
'has-[[data-flux-label]~[data-flux-control]]:grid-cols-[1fr_auto]',
'has-[[data-flux-control]~[data-flux-label]]:grid-cols-[auto_1fr]',
'[&>[data-flux-control]~[data-flux-description]]:row-start-2 [&>[data-flux-control]~[data-flux-description]]:col-start-2',
'[&>[data-flux-control]~[data-flux-error]]:col-span-2 [&>[data-flux-control]~[data-flux-error]]:mt-1', // Position error messages...
'[&>[data-flux-label]~[data-flux-control]]:row-start-1 [&>[data-flux-label]~[data-flux-control]]:col-start-2',
```

The thing to note is these styles are being applied based on the existence of a `data-flux-control` element inside the field.

But if we have a look at how an input is structured, we can see that while the input has `data-flux-control` on it, it's wrapped in a div that has `data-flux-input` on it.

```
<div ... data-flux-input>
    ....

    <input
        ...
        data-flux-control
        data-flux-group-target
        ...
    >

    ...
</div>
```

This means that the above classes don't apply as there is no `data-flux-control` as a direct child of the field.

# The solution

The solution is to add classes that add support for a field with a `data-flux-input` inside of it.

So this PR adds the following classes (in addition to the ones above already on the field component).

```
'has-[[data-flux-label]~[data-flux-input]]:grid-cols-[auto_1fr]',
'has-[[data-flux-input]~[data-flux-label]]:grid-cols-[1fr_auto]',
'[&>[data-flux-input]~[data-flux-description]]:row-start-2 [&>[data-flux-input]~[data-flux-description]]:col-start-2',
'[&>[data-flux-input]~[data-flux-error]]:col-span-2 [&>[data-flux-input]~[data-flux-error]]:mt-1', // Position error messages...
'[&>[data-flux-label]~[data-flux-input]]:row-start-1 [&>[data-flux-label]~[data-flux-input]]:col-start-2',
```

Now inline works as expected.

Screenshot 2025-11-25 at 01 40 44PM@2x

You might be wondering, can we not just do this all in the same set of classes? And yes, it might be possible, especially if we were to move the `data-flux-control` from the input to it's wrapping div.

But that would be a breaking change and currently the Laravel starter kits have CSS selectors targeting input with `data-flux-control` specifically. https://github.com/laravel/livewire-starter-kit/blob/7c00af994b1c87f6775011b9fa428b0d5a5cf176/resources/css/app.css#L58

There is also another problem.

In the scenario of a checkbox, the following would get applied `has-[[data-flux-label]~[data-flux-control]]:grid-cols-[1fr_auto]`.

This means that the label would take up all of the available space and the checkbox control would shrink.

If we were to apply this to an input though, it would mean the label takes up a lot of space and the input is tiny.

So the classes applied above for the input have also been tweaked to ensure they work properly with inputs, such as the grid cols being swapped `has-[[data-flux-label]~[data-flux-input]]:grid-cols-[auto_1fr]`.

Fixes livewire/flux#1784

### DIFF
--- a/stubs/resources/views/flux/field.blade.php
+++ b/stubs/resources/views/flux/field.blade.php
@@ -14,11 +14,18 @@ $classes = Flux::classes()
         default => 'block',
         'inline' => [
             'grid gap-x-3 gap-y-1.5',
+            // Support for inline controls...
             'has-[[data-flux-label]~[data-flux-control]]:grid-cols-[1fr_auto]',
             'has-[[data-flux-control]~[data-flux-label]]:grid-cols-[auto_1fr]',
             '[&>[data-flux-control]~[data-flux-description]]:row-start-2 [&>[data-flux-control]~[data-flux-description]]:col-start-2',
             '[&>[data-flux-control]~[data-flux-error]]:col-span-2 [&>[data-flux-control]~[data-flux-error]]:mt-1', // Position error messages...
             '[&>[data-flux-label]~[data-flux-control]]:row-start-1 [&>[data-flux-label]~[data-flux-control]]:col-start-2',
+            // Support for inline inputs...
+            'has-[[data-flux-label]~[data-flux-input]]:grid-cols-[auto_1fr]',
+            'has-[[data-flux-input]~[data-flux-label]]:grid-cols-[1fr_auto]',
+            '[&>[data-flux-input]~[data-flux-description]]:row-start-2 [&>[data-flux-input]~[data-flux-description]]:col-start-2',
+            '[&>[data-flux-input]~[data-flux-error]]:col-span-2 [&>[data-flux-input]~[data-flux-error]]:mt-1', // Position error messages...
+            '[&>[data-flux-label]~[data-flux-input]]:row-start-1 [&>[data-flux-label]~[data-flux-input]]:col-start-2',
         ],
     })
     ->add(match ($variant) {


calebporzio:
Hmmmm interesting, the inline field thing was really meant for things like checkboxes/radios/switches as you typically want those inline even among other normal fields.

My hesitancy here is that in reality someone putting text inputs and other full-width controls inline will want to make sure that all the inputs line up vertically so that they don't stagger based on the length of the label.

I might want to just say no on this and maybe we can update the docs somehow to express that or something?

Idk what do you think?

JamesKennewell:
From the Perspective of someone using Flux,

i would like to see this added. ive got a few forms where id like to have all the input labels down the left hand side

That being said i would also want the flux:inputs to align on the left side.

Could this be done with a new variant?

sertxudev:
> My hesitancy here is that in reality someone putting text inputs and other full-width controls inline will want to make sure that all the inputs line up vertically so that they don't stagger based on the length of the label.

I'm with you. To be able to line up vertically all the inputs, we'll need to design the layout with two columns, one for labels and another for the fields.

joshhanley:
Yeah I agree. This is something that will require a re-think of field/label/error and whatever input/control. So I'm going to close this PR.

Issue #1617, opened 2025-05-20
Routes where app url has a prefix


plgs:
### Flux version

v2.1.6

### Livewire version

v3.6.3

### Tailwind version

v4.1.7

### Browser and Operating System

Chrome on Windows

### What is the problem?

I'm using the Laravel Livewire Starter Kit.  For various reasons, my app URI has a prefix, so routes look like
`/myprefix/register`
`/myprefix/dashboard` etc

I can get Laravel to recognise this by using
`APP_URL=http://mydomain.com/myprefix`
in my Laravel .env file.

I can also get Livewire to recognise the prefix by using
```php
Livewire::setUpdateRoute(function ($handle) {
        return Route::post('/myprefix/livewire/update', $handle);
        });
Livewire::setScriptRoute(function ($handle) {
        return Route::get('/myprefix/livewire/livewire.js', $handle);
        });
```
in AppServiceProvider::boot() in file AppServiceProvider.php

However, I can't get Flux to use other than its top-level route
`/flux/flux.js`
which is getting a 404
`"GET /flux/flux.js?id=3ec2d92c HTTP/1.1" 404`
and I'm guessing this is why Flux controls aren't working (appear but are unresponsive), even in the Starter Kit app shell.

Is there a way to get Flux to use a prefix in its routes?

(I've seen the note on the Flux installation page about "Configuring nginx" but I'm not using nginx.  Is there a solve like this for Apache httpd?).

Thanks.

### Code snippets

As above.

### How do you expect it to work?

Some way of configuring Flux to use routes with a prefix.

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

plgs:
Thanks so much - this has worked for me to correct flux routing for my setup.

(Still have a problem finding flux.js itself, but I think that's a separate issue).

Much appreciated.

joshhanley:
@plgs you shouldn't update vendor files manually, that's a bad practice. I'm going to reopen this as that isn't the solution.

Jaiieq:
@joshhanley I 100% agree that this is bad practice, but unless it is undocumented, I know of no other way to get around the issue.  As I have noted with Flux, to my knowledge, has no methods like Livewire to adjust the endpoints, but would LOVE for that to be available!

[...]

PR #2183, opened 2025-11-27
Add support for assets in subfolders

joshhanley:
# The scenario

Currently if you run Laravel from a subfolder, like `http://localhost:8080/test-site/`, Flux's assets return a 404.

Screenshot 2025-11-27 at 12 49 57PM@2x

# The problem

The problem is that Flux's `<script>` and `<link>` tags reference a root-relative URL. This means that when rendered they look like:

```html
<script src="/flux/flux.js?id=6bd8aa7b" data-navigate-once=""></script>
```

So this means that it's actually looking for Flux's javascript in `localhost:8080/flux/flux.js`.

As you can see though, the subfolder of `test-site` isn't there and as that's where the Laravel app is being served from, it's returning a 404.

# The solution

The solution is actually quite simple, we can wrap the `/flux/flux.js` in the `url()` helper, like `url('/flux/flux.js')` which will actually give us an absolute URL including the domain name and any subfolders.

This PR adds the url helper to all of Flux's asset paths.

You can see in the screenshot below that I tested this on a standard setup, subdomains and subfolders and all are working happily.

**Standard install**
Screenshot 2025-11-27 at 12 43 33PM@2x

**Subdomain**
Screenshot 2025-11-27 at 12 43 12PM@2x

**Subfolder**
Screenshot 2025-11-27 at 12 41 11PM@2x

I also checked this against the problems we have had with Livewire's assets/urls in the past (localisation, tenants, etc.) and the primary issue when its comes to localisation/tenants is Livewire's update route. The actual asset URLs don't need to be localised/tenanted from what I could find.

Because of that, I believe that this fix should be sufficient for Flux, as it only has assets and no update route.

Fixes livewire/flux#1617

### DIFF
--- a/src/AssetManager.php
+++ b/src/AssetManager.php
@@ -82,9 +82,9 @@ class AssetManager
         $nonce = isset($options) && isset($options['nonce']) ? ' nonce="' . $options['nonce'] . '"' : '';
 
         if (config('app.debug')) {
-            return '<script src="/flux/flux.js?id='. $versionHash . '" data-navigate-once' . $nonce . '></script>';
+            return '<script src="'. url('/flux/flux.js?id='. $versionHash) . '" data-navigate-once' . $nonce . '></script>';
         } else {
-            return '<script src="/flux/flux.min.js?id='. $versionHash . '" data-navigate-once' . $nonce . '></script>';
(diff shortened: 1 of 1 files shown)


calebporzio:
Hell yeah, super simple, love it.

Issue #2199, opened 2025-12-01
OTP input group and private properties don't work together


DucPanigale:
### Flux version

v2.6

### Livewire version

v3.7.0

### Tailwind version

v4.0.7

### Browser and Operating System

Edge on Windows 11

### What is the problem?

Hi y'all!

I love the new OTP input but sadly there seems to be a small issue.
When adding the property "private" to a OTP input group it does not mask the input. If it's not a group, the input is masked.

As a heads up: even though it is in the manual before creating a bug report - I don't use @fluxAppearance because I don't want to use the dark mode.

I hope you can help me. 
Kind regards

### Code snippets to replicate the problem

<flux:otp wire:model="pin" private>
	<flux:otp.group>
		<flux:otp.input />
		<flux:otp.input />
	</flux:otp.group>
</flux:otp>

### Screenshots/ screen recordings of the problem

<img width="99" height="73" alt="Image" src="https://github.com/user-attachments/assets/de231e9d-51bd-40ea-a1b8-6b71e5d4b7aa" />

### How do you expect it to work?

When adding the property "private" the password should be masked.

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

ganyicz:
Thanks for reporting! I've submitted a PR for this, should be fixed in the next release.

As a temporary solution, you can manually add `type="password"` to the inputs like so:

```blade
<flux:otp label="PIN Code">
    <flux:otp.group>
        <flux:otp.input type="password" />
        <flux:otp.input type="password" />
        <flux:otp.input type="password" />
        <flux:otp.input type="password" />
    </flux:otp.group>
</flux:otp>
```

PR #2202, opened 2025-12-01
Fix private OTP within a group

ganyicz:
# The scenario

An OTP with manually defined inputs inside a group doesn't respect the `private` prop:

```
<flux:otp private label="PIN Code">
    <flux:otp.group>
        <flux:otp.input />
        <flux:otp.input />
        <flux:otp.input />
        <flux:otp.input />
    </flux:otp.group>
</flux:otp>
```

These will be displayed as regular text input without masking.

# The problem

The `private` prop only works for inputs rendered from inside the main `flux:otp` component (when `length` is used).

# The solution

Moving the `private` logic from the main component to the individual input components + `@aware` fixes this

Fixes livewire/flux#2199


calebporzio:
great

Issue #2209, opened 2025-12-04
Time picker missing from flux:publish


skeemer:
### Flux version

v2.9.1

### Livewire version

v4.0.0-beta.3

### Tailwind version

4.1.17

### Browser and Operating System

Brave on Pop!_OS

### What is the problem?

I can't publish the time picker via the artisan command. It's in the stubs folder and copying that worked fine so far.

### Code snippets to replicate the problem

```shell
./artisan flux:publish
```
Look for the time picker.

### Screenshots/ screen recordings of the problem

<img width="663" height="205" alt="Image" src="https://github.com/user-attachments/assets/2b33b30b-727e-49a7-869b-0e5784cf6a0f" />

### How do you expect it to work?

Select "Time picker".

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

joshhanley:
@skeemer thanks for reporting! This has been fixed and will be in the next release.

PR #2211, opened 2025-12-04
Update publish command

ganyicz:
Updates the publish command with all recently added components.

Fixes livewire/flux#2209


calebporzio:
ooof good catch thanks

Issue #2234, opened 2025-12-12
Input Loading state always active with wire:model.live


gRoberts84:
### Flux version

v2.9.2

### Livewire version

v3.7.1

### Tailwind version

v4.1.11

### Browser and Operating System

Chrome on macOS

### What is the problem?

Using wire:model.live on `<flux:input />` causes loader to appear all the time.

### Code snippets to replicate the problem

```
<?php
namespace App\Livewire\Components;

use Livewire\Component;
use Livewire\WithFileUploads;

class Documents extends Component
{
    use WithFileUploads;

    public $name;

    public function render()
    {
        return view('livewire.components.documents');
    }
}
```

```
<flux:input
    name="name"
    :label="__('Name')"
    type="text"
    wire:model.live="name"
    required
    autofocus
    autocomplete="name"
    :placeholder="__('Name')"
/>
```

### Screenshots/ screen recordings of the problem

https://github.com/user-attachments/assets/21c6554c-059a-4ae7-809a-8d6f8a1f9ecb

### How do you expect it to work?

Loader should only show when communicating with the server.

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

jeffchown:
@gRoberts84 Does the same thing happen for you when not in a modal?

jeffchown:
@gRoberts84 Testing your snippet via a Volt component, and not in a modal, it works OK for me:

```blade
<?php
use Livewire\Volt\Component;
use Livewire\WithFileUploads;

new class extends Component
{
    use WithFileUploads;

    public $name;
};
?>

<flux:input
    name="name"
    :label="__('Name')"
    type="text"
    wire:model.live="name"
    required
    autofocus
    autocomplete="name"
    :placeholder="__('Name')"
/>
```

ganyicz:
Hi @gRoberts84,

I wasn't able to replicate your issue with your example code:

https://github.com/user-attachments/assets/50feead4-d2ff-428a-add1-95aaa5db6b9d

Here is the code I used in the above video:

```blade
<?php

use Livewire\Component;
use Livewire\WithFileUploads;

new class extends Component
{
    use WithFileUploads;

    public string $name = '';
};
?>

<div class="w-56">
    <flux:input
        name="name"
        :label="__('Name')"
        type="text"
[...]

PR #2235, opened 2025-12-12
Fixed bug with input loading state that caused it to always show load…

gRoberts84:
…ing with wire:model.live

# The scenario

When using `wire:model.live`, the loading state is always active.

# The problem

Livewire and/or Flux was not correctly hiding the loading state when not needed.

# The solution

Added default hidden class and updated the `wire:loading` property to remove the class when busy.

Fixes livewire/flux#


ganyicz:
Hi @gRoberts84,

I replied to your issue (https://github.com/livewire/flux/issues/2234) regarding this.

I wasn't able to replicate this, can you please reply or update the issue description with an example code that reproduces the error? Thanks!

ganyicz:
I'm closing this as the original issue was also closed so I assume this is not needed anyomore. Thanks!

PR #2236, opened 2025-12-12
Only activate the placeholder active styles when not focused.

gRoberts84:
# The scenario

Whilst this likely affects both light and dark mode, it was most noticable in Light mode, where `has-[option.placeholder:checked]:text-zinc-400` would cause all options to appear as if they were disabled.

# The problem

Applied styles were active as long as the placeholder was selected, causing all other options to appear as if they were placeholders.

# The solution

Updated the applied styles to only activate when the select is not focused. When open (focused) the browser will handle the presentation of the disabled placeholder, allowing the options to be visible as they should be.

Fixes livewire/flux#


ganyicz:
Hi @gRoberts84, thanks for the PR!

Can you please update the description with before/after screenshots of the change and also an example code to replicate the issue you're describing?

Thanks!

ganyicz:
@gRoberts84 

I'm going to close this now as I couldn't replicate the bug, however feel free to open a separate issue with screenshots/video and we'll look into it!

Issue #2250, opened 2025-12-17
Flux uses wire:current.ignore (Livewire v4) causing issues on Livewire v3


NimaKhadivar:
### Flux version

2.10.0

### Livewire version

livewire 3

### Tailwind version

4.0.0

### Browser and Operating System

chrome on macos

### What is the problem?

all sidebar nav item are active

### Code snippets to replicate the problem

-

### Screenshots/ screen recordings of the problem

<img width="1200" height="767" alt="Image" src="https://github.com/user-attachments/assets/2773c294-d803-485c-a567-16abbdaf1253" />

<img width="1351" height="774" alt="Image" src="https://github.com/user-attachments/assets/88d506d7-33fc-4e59-b032-e815e8f431b4" />

### How do you expect it to work?

only 1 of them be active

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

abishekrsrikaanth:
@NimaKhadivar  @calebporzio, the `wire:current.ignore` was released as part of `v4 beta release` (https://github.com/livewire/livewire/pull/9655) but this PR https://github.com/livewire/flux/pull/2246 makes use of this attribute and is released as part of the latest flux release causing the issue.

ganyicz:
@NimaKhadivar @abishekrsrikaanth 

Thanks for reporting! 

I submitted PRs to both Flux and Livewire with fixes.

PR #2258, opened 2025-12-17
Fix wire:current.ignore

ganyicz:
# The scenario

When using Livewire v3 or Livewire v4-beta.4 all links show up as active:

SCR-20251217-nnel

# The problem

In #2246, we added `wire:current.ignore` when `:current` is manually set on a link.

However because of incorrect implementation, the `wire:current.ignore` attribute is being added to all links.

This is what the previous implementation looked like:

```blade
{{ $attributes->merge(['data-current' => $current, 'wire:current.ignore' => $current !== null]) }}
```

But by this time `$current` won't be `null` because of code above this line:

```php
$current = $current === null ? ($hrefForCurrentDetection
    ? $requestIs($hrefForCurrentDetection === '/' ? '/' : trim($hrefForCurrentDetection, '/'))
    : false) : $current;
```

And because previous Livewire versions don't support for this attribute, it acts as `wire:current` without any modifier.

# The solution

I submitted a PR to backport this attribute to v3: https://github.com/livewire/livewire/pull/9667.

_We should also bump Livewire version requirement in Flux after release._

To prevent the attribute from being added to all links, I refactored the code so that the attribute is only added when `$current` is actually specified:

```blade
@php
if ($as !== 'div' || $href) {
    if ($current !== null) {
        // If the user manually specified :current="true/false", we need to stop Livewire from managing
        // the data-current attribute as it would be automatically added/removed when using wire:navigate...
        $attributes = $attributes->merge(['data-current' => $current, 'wire:current.ignore' => true]);
    } else {
        ...

        $current = $hrefForCurrentDetection
            ? $requestIs($hrefForCurrentDetection === '/' ? '/' : trim($hrefForCurrentDetection, '/'))
            : false;

        $attributes = $attributes->merge(['data-current' => $current]);
    }
}
@endphp

<?php if ($as === 'div' && ! $href): ?>
    <div {{ $attributes }}>
        {{ $slot }}
    </div>
<?php elseif ($as === 'a' || $href): ?>
    <a href="{!! e($href) !!}" {{ $attributes }}>
        {{ $slot }}
    </a>
<?php else: ?>
    <button {{ $attributes->merge(['type' => $type]) }}>
        {{ $slot }}
    </button>
<?php endif; ?>
```

Fixes livewire/flux#2250

### DIFF
--- a/stubs/resources/views/flux/button-or-link.blade.php
+++ b/stubs/resources/views/flux/button-or-link.blade.php
@@ -15,22 +15,32 @@ extract(Flux::forwardedAttributes($attributes, [
 ])
 
 @php
-$hrefForCurrentDetection = str($href)->startsWith(trim(config('app.url')))
-    ? (string) str($href)->after(trim(config('app.url'), '/'))
-    : $href;
+if ($as !== 'div' || $href) {
+    if ($current !== null) {
+        // If the user manually specified :current="true/false", we need to stop Livewire from managing
+        // the data-current attribute as it would be automatically added/removed when using wire:navigate...
+        $attributes = $attributes->merge(['data-current' => $current, 'wire:current.ignore' => true]);
+    } else {
+        $hrefForCurrentDetection = str($href)->startsWith(trim(config('app.url')))
+            ? (string) str($href)->after(trim(config('app.url'), '/'))
+            : $href;
 
-if ($hrefForCurrentDetection === '') $hrefForCurrentDetection = '/';
+        if ($hrefForCurrentDetection === '') $hrefForCurrentDetection = '/';
 
-$requestIs = function ($pattern) {
-    // Support current route detection during Livewire update requests as well...
-    return app('livewire')?->isLivewireRequest()
-        ? str()->is($pattern, app('livewire')->originalPath())
-        : request()->is($pattern);
-};
+        $requestIs = function ($pattern) {
+            // Support current route detection during Livewire update requests as well...
+            return app('livewire')?->isLivewireRequest()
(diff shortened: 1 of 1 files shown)


joshhanley:
Marked as draft as it's waiting on PR #2260 to be merged again.

joshhanley:
@ganyicz was looking at this, I maybe be misunderstanding, won't this break the built in Flux automatic current detection? At the moment, if no `current` prop is supplied, Flux will automatically do a route match with the href and add `data-current` but in your change there is no `wire:current.ignore` in that else condition, so Livewire v4 will still remove the manually added `data-current` right?

ganyicz:
Livewire v4 will only remove it if you also use wire:navigate (essentially overriding Flux), if you don't, Flux automatic detection will work the same as before:

SCR-20251218-enpt

```php
Route::view('/', 'welcome');
Route::view('/admin', 'welcome');
```

```blade
<body class="bg-zinc-900 pt-32" x-data>
    <div class="max-w-sm mx-auto">
        <flux:text class="pl-3 mb-4">URL: {{ str()->start(request()->path(), '/') }}</flux:text>
        <flux:sidebar.item href="/admin" :current="null">no current, no wire:navigate</flux:sidebar.item>
        <flux:sidebar.item href="/admin" :current="null" wire:navigate>no current, wire:navigate</flux:sidebar.item>
        <flux:sidebar.item href="/admin" :current="true" wire:navigate>current, wire:navigate</flux:sidebar.item>
    </div>

    <?php Livewire::forceAssetInjection(); ?>
</body>
```

PR #2260, opened 2025-12-18
Restore "Allow forcing of `data-current` in links"

joshhanley:
This restores PR #2246

Reverts livewire/flux#2259


joshhanley:
We need to tag a version of Livewire v3 and bump the minimum requirements in composer.json before this can be merged.

Issue #2155, opened 2025-11-23
CSP Safe Mode


azizabbas:
### Flux version

v2

### Livewire version

v3

### Tailwind version

v4

### Browser and Operating System

Chrome on MacOS

### What is the problem?

I am creating this issue as followup of feature request https://github.com/livewire/flux/discussions/2119

Flux UI generates inline Alpine.js with let statements (like let input = $el.closest(...)), which CSP-safe mode doesn't support.

This means we cannot have CSP Safe mode enabled in our livewire config if we are to use Flux UI at full, like using clear button to clear input and related.

This must be considered in future releases of Flux and maybe there is a workaround to fix this.

Currently, we have to use AlpineJS with Javascript to achieve all that if we want to have CSP Safe mode enabled.

### Code snippets to replicate the problem

-

### Screenshots/ screen recordings of the problem

-

### How do you expect it to work?

-

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

calebporzio:
Ah good point. Let's start by maybe @joshhanley  listing out all the places we use csp-violating expressions here and then we can talk next steps. I don't see why we wouldn't be able to make flux csp-safe.

joshhanley:
I've done a search to see where we use Alpine and this is what I came up with.

## Flux
**Input**
- input/clearable.blade.php
- input/copyable.blade.php
- input/expandable.blade.php
- input/file.blade.php
- input/index.blade.php
- input/viewable.blade.php

**Modal**
- modal/index.blade.php
- modal/trigger.blade.php

**Sidebar**
- sidebar/index.blade.php
- sidebar/toggle.blade.php

## Flux Pro
**Accordion**
- accordion/content.blade.php
- accordion/item.blade.php

**Command**
- command/input.blade.php

**Date Picker**
- date-picker/button.blade.php

**Pillbox**
- pillbox/search.blade.php
- pillbox/trigger.blade.php

**Select**
- select/button.blade.php
- select/input.blade.php
- select/search.blade.php

**Time Picker**
- time-picker/button.blade.php
- time-picker/input.blade.php

**Toast**
- toast/group.blade.php
- toast/index.blade.php

ganyicz:
I've submitted a PR to make all components CSP-compatible.

PR #2277, opened 2025-12-25
Make components CSP safe

ganyicz:
# The scenario

Some Flux components are not compatible with Livewire's new CSP mode.

# The problem

In some places we're using unsupported syntax such as assignments, complex conditionals, etc..

```html
x-on:click="let input = $el.closest('[data-flux-input]').querySelector('input');"
```

```html
x-on:modal-show.document="if ($event.detail.name === @js($name) && (! $event.detail.scope)) $el.showModal()"
```

# The solution

Refactoring complex code into `Alpine.data` included in compiled js files.

I've prefixed all datas with `flux` to avoid conflicts, eg.: `fluxInputClearable`, `fluxInputViewable`

Fixes livewire/flux#2155

### DIFF
--- a/stubs/resources/views/flux/input/clearable.blade.php
+++ b/stubs/resources/views/flux/input/clearable.blade.php
@@ -12,8 +12,8 @@ $attributes = $attributes->merge([
 <flux:button
     :$attributes
     :size="$size === 'sm' || $size === 'xs' ? 'xs' : 'sm'"
-    x-data
-    x-on:click="let input = $el.closest('[data-flux-input]').querySelector('input'); input.value = ''; input.dispatchEvent(new Event('input', { bubbles: false })); input.dispatchEvent(new Event('change', { bubbles: false })); input.focus()"
+    x-data="fluxInputClearable"
+    x-on:click="clear()"
     tabindex="-1"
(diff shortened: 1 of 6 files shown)


calebporzio:
make changes from pro repo first

maikezan:
Hi guys, i was succesfully using the modal (in flyout mode) for a "first-access-tutorial" for users. Since this guide needs to spreads across different sections i was including it in a layout file.

I've updated to this version today and now i get this error:
```
ErrorException
vendor/livewire/flux/stubs/resources/views/flux/modal/index.blade.php:93
Undefined variable $__livewire
```
I see there has been some changes around that code block (line 93) in this commit that removed an isset check for $__livewire variable. 
I've also noticed that moving my flux:modal code block out of the layout files fixes the error, but still unsure if the issue is mine or if it may be related to the above said change.
If it can help, i'm also using Livewire 4.0.2 at the moment.

ganyicz:
Hi @maikezan,

There is a bug in the latest Flux release with using modals outside Livewire components.

This will be fixed in the next release in couple of hours.

Until then, you can either downgrade to v2.10.0 or put your modal inside a Livewire component.

Issue #2314, opened 2026-01-15
Inconsistent icon behavior on inputs inside a flex div


dev-idkwhoami:
### Flux version

v2.10.2

### Livewire version

v4.0.0

### Tailwind version

v4.1.11

### Browser and Operating System

Brave on Windows 11

### What is the problem?

Whenever an input is aligned with a bigger element inside a div with `display: flex` , the icons of that input don't behave correct. Unfortunately the absolute position to overlay the icon causes the icon to "scale" in the wrong context.

A "simple" workaround is to wrap the input in div, however that seems like more of a band-aid than a solution.

### Code snippets to replicate the problem

```html
<div class="relative h-full flex-1 p-12 space-y-6 overflow-hidden rounded-xl border border-neutral-200 dark:border-neutral-700">
            <div class="px-12 space-x-3 w-full">
                <flux:input clearable placeholder="base no button" class="max-w-xs" icon:trailing="magnifying-glass"></flux:input>
            </div>
            <div class="flex px-12 space-x-3 w-full">
                <flux:input clearable placeholder="base with base button" class="max-w-xs" icon:trailing="magnifying-glass"></flux:input>
                <flux:button variant="primary"></flux:button>
            </div>
            <div class="px-12 space-x-3 w-full">
                <flux:input clearable placeholder="sm no button" class="max-w-xs" size="sm" icon:trailing="magnifying-glass"></flux:input>
            </div>
            <div class="flex px-12 space-x-3 w-full">
                <flux:input clearable placeholder="sm with base button" class="max-w-xs" size="sm" icon:trailing="magnifying-glass"></flux:input>
                <flux:button variant="primary"></flux:button>
            </div>
            <div class="px-12 space-x-3 w-full">
                <flux:input clearable placeholder="xs no button" class="max-w-xs" size="xs" icon:trailing="magnifying-glass"></flux:input>
            </div>
            <div class="flex px-12 space-x-3 w-full">
                <flux:input clearable placeholder="xs with base button" class="max-w-xs" size="xs" icon:trailing="magnifying-glass"></flux:input>
                <flux:button variant="primary"></flux:button>
            </div>
            <div class="flex px-12 space-x-3 w-full">
                <div class="w-full max-w-xs">
                    <flux:input clearable placeholder="wrapped xs with base button"  size="xs" icon:trailing="magnifying-glass"></flux:input>
                </div>
                <flux:button variant="primary"></flux:button>
            </div>
[...]

PR #2315, opened 2026-01-15
Add height to input wrapper

ganyicz:
# The scenario

When input's height is stretched beyond its original height, its icon becomes misaligned:

```blade
<div class="flex">
    <flux:input size="xs" icon:trailing="magnifying-glass"></flux:input>
    <flux:button variant="primary">Search</flux:button>
</div>
```

SCR-20260119-paxn

# The problem

This happens because the input has a fixed height:

```php
$classes = Flux::classes()
    ->add(match ($size) {
        default => '... h-10 ...',
        'sm' => '... h-8 ...',
        'xs' => '... h-6 ...',
    })
```

However the icon is absolutely positioned inside input wrapper with auto height:

```blade
<div>
    <input class="{{ $classes }}">
    
    <div class="absolute top-0 bottom-0 flex items-center">
        <flux:icon :icon="$iconTrailing" />
    </div>
</div>
```

# The solution

Is to add matching fixed height to the wrapper:

```php
$wrapperClasses = Flux::classes()
    ->add(match ($size) {
        default => 'h-10',
        'sm' => 'h-8',
        'xs' => 'h-6',
    })
    ;
```

```blade
<div class="{{ $wrapperClasses }}">
    <input class="{{ $classes }}">
    
    ...
</div>
```

Fixes livewire/flux#2314

### DIFF
--- a/stubs/resources/views/flux/input/index.blade.php
+++ b/stubs/resources/views/flux/input/index.blade.php
@@ -119,6 +119,15 @@ $classes = Flux::classes()
     })
     ->add($attributes->pluck('class:input'))
     ;
+
+$wrapperClasses = Flux::classes()
+    ->add('w-full relative block group/input')
+    ->add(match ($size) {
+        default => 'h-10',
+        'sm' => 'h-8',
+        'xs' => 'h-6',
+    })
+    ;
 @endphp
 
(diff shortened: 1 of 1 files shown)


calebporzio:
gonna pass on this for now tyty

Issue #2300, opened 2026-01-09
Wrapped listbox in custom component inherits option label


ju5t:
### Flux version

v2.10.2

### Livewire version

v3.7.3

### Tailwind version

v4

### Browser and Operating System

Chrome on MacOS

### What is the problem?

We have a custom component that wraps a `flux:select`. It uses the `listbox` variant. 

This component has a `label` property that somehow gets inherited by the `flux:select.option` component. As a result, even though we set the `$slot`, the label gets preference over it. 

```blade
<flux:select.option value="{{ $option['id'] }}" wire:key="{{ $option['id'] }}">
    <div class="flex items-center gap-2">
        {{ $option['name'] }}
    </div>
</flux:select.option>
```

I did some digging and found that even though we don't specify any variant on the select option, it doesn't use the default variant, but switches to the custom one.

The custom variant has:

```blade
{{ $label ?? $slot }}
```

Which results in the label of the parent component being used for each option instead of what we pass in its slot.

See also:
https://github.com/livewire/flux/issues/1847

### Code snippets to replicate the problem

```bash
php artisan make:view custom-component
```

And paste this in the blade file:

```blade
@props([
    'for',
    'label',
    'options' => [],
])

@php
    $listbox = collect($options)
        ->values()
        ->toArray();
@endphp

<flux:select {{ $attributes->whereDoesntStartWith('class') }} size="sm" variant="listbox">
    @foreach ($listbox as $option)
        <flux:select.option value="{{ $option['id'] }}" wire:key="{{ $option['id'] }}">
            <div class="flex items-center gap-2">
                {{ $option['name'] }}
            </div>
        </flux:select.option>
    @endforeach
</flux:select>
```

Example usage:

```blade
@php
$options = [
    ['id' => 'foo', 'name' => 'bar',],
];
@endphp

<x-custom-component :options="$options" />
```

### Screenshots/ screen recordings of the problem

n/a

### How do you expect it to work?

I would expect the slot to work and that it would not inherit the label from the parent component.

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

ganyicz:
Hi @ju5t,

thanks for reporting!

I'll just leave a note for myself here:

[...]

PR #2371, opened 2026-02-05
Fix label/slot precedence

ganyicz:
# The scenario

A user-land component with a `label` prop renders select options incorrectly:

```blade
<x-select label="Industry" />
```

```blade
{{-- resources/views/components/select.blade.php --}}
@props(['label'])

<flux:select :$label variant="listbox">
</flux:select>
```

SCR-20260209-nimv-2

# The problem

The `label` prop leaks from the parent component into all children.

This happens because `<flux:delegate-component>` forwards inherited data from the entire component stack.

Combined with `{{ $label ?? $slot }}`, every option displays "Industry".

# The solution

Reverse the priority from `{{ $label ?? $slot }}` to `{{ $slot->isNotEmpty() ? $slot : $label }}` across 9 files where this pattern occurs. Slot content now wins when present; `label` prop still works as fallback for self-closing tags.

# Risk assessment

* The `$slot->isNotEmpty()` pattern is already used elsewhere in the codebase
* The docs never show both being passed simultaneously
* Self-closing tags (`<flux:radio label="Foo" />`) produce an empty slot, so `$label` is still used as fallback

The problematic pattern would be:

```blade
<flux:menu.radio label="Foo">Bar</flux:menu.radio>
```

Before this change: Renders `Foo`
After this change: Renders `Bar`

Flux Pro PR: livewire/flux-pro#436

Fixes livewire/flux#2300

### DIFF
--- a/stubs/resources/views/flux/checkbox/variants/buttons.blade.php
+++ b/stubs/resources/views/flux/checkbox/variants/buttons.blade.php
@@ -61,7 +61,7 @@ $iconAttributes = Flux::attributesAfter('icon:', $attributes, [
         {{ $icon }}
     <?php endif; ?>
 
-    @if (isset($label) || $slot->isNotEmpty())
-        <span>{{ $label ?? $slot }}</span>
+    @if ($slot->isNotEmpty() || isset($label))
+        <span>{{ $slot->isNotEmpty() ? $slot : $label }}</span>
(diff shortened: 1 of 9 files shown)


ju5t:
Somehow didn't see this one, awesome stuff @ganyicz!

PR #2401, opened 2026-02-16
Add scroll-to prop to pagination component

joshhanley:
# The Scenario

When using Flux's pagination component, clicking a pagination button changes the page but doesn't scroll the user back to the top. This is a common expectation — especially on long lists where the pagination controls sit at the bottom of the page — and is something Livewire's own pagination views support out of the box.

```blade
<?php

use App\Models\User;
use Livewire\Attributes\Computed;
use Livewire\Component;
use Livewire\WithPagination;

new class extends Component {
    use WithPagination;

    #[Computed]
    public function users()
    {
        return User::query()->paginate(10);
    }
}; ?>

<div>
    <flux:table :paginate="$this->users">
        <flux:table.columns>
            <flux:table.column>Name</flux:table.column>
            <flux:table.column>Email</flux:table.column>
        </flux:table.columns>

        <flux:table.rows>
            @foreach ($this->users as $user)
                <flux:table.row :key="$user->id">
                    <flux:table.cell>{{ $user->name }}</flux:table.cell>
                    <flux:table.cell>{{ $user->email }}</flux:table.cell>
                </flux:table.row>
            @endforeach
        </flux:table.rows>
    </flux:table>
</div>
```

# The Problem

The `<flux:pagination>` component had no scroll behaviour implementation. Livewire's pagination views include a `scrollIntoView()` call on each button via `x-on:click`, but Flux's pagination component didn't have an equivalent.

# The Solution

Added an opt-in `scroll-to` prop to the `<flux:pagination>` component, matching the approach used by Livewire's pagination views. When enabled, clicking a pagination button triggers a `scrollIntoView()` call on the target element.

The prop is **opt-in** (disabled by default) and accepts either a boolean or a CSS selector:

```blade
{{-- Scroll to body (default target) --}}
<flux:pagination :paginator="$users" scroll-to />

{{-- Scroll to a specific element --}}
<flux:pagination :paginator="$users" scroll-to="#table" />
```

When using the table component, the prop can be forwarded using the `pagination:` prefix:

```blade
<flux:table :paginate="$users" pagination:scroll-to />
```

The table component now uses `Flux::attributesAfter('pagination:', ...)` to forward any `pagination:`-prefixed attributes to the underlying `<flux:pagination>` component, making it extensible for future pagination props.

**Note:** There is a known issue in Laravel where multi-word kebab-case attributes (like `scroll-to`) don't get converted to their camelCase prop equivalents (like `$scrollTo`) when forwarded via `:attributes`. To work around this, the pagination component uses `Flux::forwardedAttributes()` before `@props` to manually handle the conversion — the same pattern used by the `with-field` component.

Fixes livewire/flux#1596

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,15 +11,15 @@
     ],
     "require": {
         "php": "^8.1",
-        "illuminate/console": "^10.0|^11.0|^12.0",
-        "illuminate/support": "^10.0|^11.0|^12.0",
-        "illuminate/view": "^10.0|^11.0|^12.0",
-        "symfony/console": "^6.0|^7.0",
+        "illuminate/console": "^10.0|^11.0|^12.0|^13.0",
+        "illuminate/support": "^10.0|^11.0|^12.0|^13.0",
+        "illuminate/view": "^10.0|^11.0|^12.0|^13.0",
+        "symfony/console": "^6.0|^7.0|^8.0",
         "livewire/livewire": "^3.7.4|^4.0",
         "laravel/prompts": "^0.1|^0.2|^0.3"
     },
     "conflict": {
-        "livewire/blaze": "<1.0.0"
+        "livewire/blaze": "<1.0.0-beta.2"
     },
     "autoload": {
         "psr-4": {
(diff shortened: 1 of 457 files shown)


calebporzio:
resolve conflicts, then good to merge 👌

PR #2449, opened 2026-03-01
Add toggle prop to OTP for private mode visibility

aydinza:
## Summary

- Adds a `toggle` prop to `flux:otp` that renders an eye/eye-slash button to toggle input visibility when used alongside `private`
- Uses Alpine's `x-effect` to reactively switch input types between `password` and `text`
- Only renders the toggle button when both `private` and `toggle` are set

## Usage

```blade
<flux:otp wire:model="pin" length="6" private toggle />
```

Also works with custom slot markup:

```blade
<flux:otp wire:model="pin" private toggle>
    <flux:otp.group>
        <flux:otp.input />
        <flux:otp.input />
        <flux:otp.input />
        <flux:otp.input />
        <flux:otp.input />
        <flux:otp.input />
    </flux:otp.group>
</flux:otp>
```

## Motivation

PIN and OTP inputs often need a visibility toggle for accessibility and usability. Currently, achieving this requires custom Alpine code outside the component. This prop provides a native, zero-config solution.

### DIFF
--- a/stubs/resources/views/flux/otp/index.blade.php
+++ b/stubs/resources/views/flux/otp/index.blade.php
@@ -3,6 +3,7 @@
 @props([
     'length' => null,
     'private' => false,
+    'toggle' => false,
 ])
 
 @php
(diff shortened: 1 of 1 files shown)


ganyicz:
Hi @aydinza 

Thanks for contributing. We're going to hold off on this for now. I see the need for this however I'm unsure about the API, design and implementation. If you believe this should be part of Flux please open a discussion with a feature request and we will consider this once there's enough demand from community.

PR #2452, opened 2026-03-02
Fix `flux:switch` indicator position with custom base font-size

### LINKED ISSUE #2443: flux:switch doesn't scale properly with different root font-size
### Flux version

v2.12.1

### Livewire version

v4.1.4

### Tailwind version

v4.2.1

Chrome on Windows

### What is the problem?

The `flux:switch` component looks bad if you change the root font-size from its default `16px` to something else.
Try changing the font; the switch doesn't scale properly and the inner circle will either go too far or too short.

```html
<html style="font-size:20px">
...

<flux:switch />
...
</html>
```

.

I think it should scale properly.

joshhanley:
# The Scenario

When the base `font-size` is changed from the default `16px`, the `flux:switch` indicator circle doesn't position correctly. It either overshoots or undershoots its target position depending on whether the font-size is larger or smaller than `16px`.

Screenshot 2026-03-02 at 02 38 29PM

```blade
<html style="font-size: 20px">

<flux:switch />

</html>
```

# The Problem

The switch container and indicator use rem-based sizing (`w-8` = 2rem, `h-5` = 1.25rem, `size-3.5` = 0.875rem), but the translate values that position the indicator use fixed pixel values (`translate-x-[3px]`, `translate-x-[2px]`, `translate-x-[15px]`).

When the base font-size changes, the container and indicator scale proportionally (since they use rem), but the translate offsets stay fixed (since they use px), causing a mismatch.

# The Solution

Convert the pixel-based translate values to their rem equivalents so they scale proportionally with the rest of the component:

- `3px` → `0.1875rem`
- `2px` → `0.125rem`
- `15px` → `0.9375rem`

Screenshot 2026-03-02 at 02 35 12PM

Fixes #2443

### DIFF
--- a/stubs/resources/views/flux/switch.blade.php
+++ b/stubs/resources/views/flux/switch.blade.php
@@ -29,13 +29,13 @@ $classes = Flux::classes()
 $indicatorClasses = Flux::classes()
     ->add('size-3.5')
     ->add('rounded-full')
-    ->add('transition translate-x-[3px] dark:translate-x-[2px] rtl:-translate-x-[3px] dark:rtl:-translate-x-[2px]')
+    ->add('transition translate-x-[0.1875rem] dark:translate-x-[0.125rem] rtl:-translate-x-[0.1875rem] dark:rtl:-translate-x-[0.125rem]')
     ->add('bg-white')
     ->add([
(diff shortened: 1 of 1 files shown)


calebporzio:
tyty

Issue #2508, opened 2026-03-21
Timeline content overflows container with long unbreakable content


schoberg:
### Flux version

v2.13.0

### Livewire version

v3.7.11

### Tailwind version

v4.1.18

### Browser and Operating System

Chrome on MacOS

### What is the problem?

`flux:timeline.content` doesn't constrain its width when containing long unbreakable content (like `<pre>` code blocks with URLs or larger images, etc). The content pushes the timeline wider than its parent container because the grid/flex child defaults to min-width: auto. ([known quirk](https://css-tricks.com/preventing-a-grid-blowout/))

Adding `overflow-hidden` or `min-w-0` to `<flux:timeline.content>` fixes it, but this should be set internally on the content slot element.

### Code snippets to replicate the problem

```
<flux:timeline.content>
<pre style="overflow: auto;max-width: 100%;padding: 10px 15px;font-size: 12px;border: 1px solid #ccc;border-radius: 8px;background: #fff;color: #333;margin: .25em 0;">Long pre tag that has expected max-width:100%; and overflow:auto; that breaks the width</pre>
</flux:timeline.content>
```

### Screenshots/ screen recordings of the problem

https://img.dropinblog.com/mn84N5TV 

### How do you expect it to work?

child elements should respect the container width

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

PR #2510, opened 2026-03-21
Fix timeline content overflow with long unbreakable content

ganyicz:
# The scenario

`<flux:timeline.content>` overflows its container when it contains long unbreakable content like `<pre>` blocks with URLs.

```html
<flux:timeline.content>
    <pre style="overflow: auto; max-width: 100%;">https://example.com/this-is-a-very-long-url-that-cannot-break...</pre>
</flux:timeline.content>
```

Before fix - content overflows container

# The problem

`[data-flux-timeline-content]` is a CSS grid item placed in a `1fr` column, but it doesn't override the default `min-width: auto`. For grid items, `min-width: auto` resolves to `min-content` — the item refuses to shrink below its intrinsic content width. The `1fr` track expands from ~420px to ~1500px to accommodate the unbreakable text.

# The solution

Add `min-width: 0` to `[data-flux-timeline-content]`, allowing the `1fr` track to constrain to the remaining available space.

After fix - content properly contained

Fixes livewire/flux#2508

### DIFF
--- a/dist/flux.css
+++ b/dist/flux.css
@@ -371,6 +371,7 @@ Timeline
         grid-column: 3;
         grid-row: 2 / 5;
         align-self: center;
+        min-width: 0;
     }
 
     [data-flux-timeline-line-leading] {


calebporzio:
legendary

PR #2512, opened 2026-03-21
Allow progress bar color to be controlled dynamically

### LINKED ISSUE #2487: Progress color is not reactive without Livewire re-render
### Flux version

v2.13.0

### Livewire version

v4.2.1

### Tailwind version

v4.2.1

Chrome on windows

### What is the problem?

When using flux:progress, updating the progress value from JavaScript works if we manually change the width, but changing the color does not update correctly unless the component is re-rendered by Livewire.

This becomes problematic for realtime scenarios (e.g. websocket updates) where progress updates multiple times per second and triggering Livewire re-renders is not feasible.

#### Example use case

A progress bar updated via WebSocket:

- progress value updates multiple times per second
- progress color changes depending on status (success / warning / error)
`<flux:progress
    x-bind:value="value"
    x-bind:color="color"
/>`
Updating value works correctly, but updating color does not change the visual color unless Livewire re-renders the component.

Currently:

updating width works via DOM

updating color requires Livewire re-render

Expected behavior

Color should update reactively when the color attribute changes without requiring a full Livewire re-render.

<div x-data="progressDemo()"
        x-init="start()"
        class="space-y-4 w-96">

        <flux:progress
            x-bind:value="value"
            x-bind:color="color"
        />

        <div class="text-sm text-gray-500">
            Progress: <span x-text="value"></span>%
            | Color: <span x-text="color"></span>
        </div>

    </div>

    <script>
        function progressDemo() {
            return {
                value: 0,
                color: 'green',

                colors: [
                    'red','green','blue'
                ],

                start() {
                    setInterval(() => {
                        this.value = Math.floor(Math.random() * 100)
                        this.color = this.colors[Math.floor(Math.random() * this.colors.length)]
                    }, 3000)
                }
            }
        }
    </script>

Image

Color should update reactively when the color attribute changes without requiring a full Livewire re-render.

ganyicz:
## The scenario

The progress bar color can't be changed dynamically from Alpine.

Value updates work via `x-bind:value`, but color doesn't:

```blade
<flux:progress
    x-bind:value="progress"
    x-bind:color="failed ? 'red' : 'green'"
/>
```

## The problem

The color is baked in at render time. 

```php
$barClasses = Flux::classes()
    ->add(match ($color) {
        'red'   => 'bg-red-600 dark:bg-red-400',
        'green' => 'bg-green-600 dark:bg-green-400',
        // ...
    })
    ;
```

## The solution

We expose the bar color as a CSS variable (`--flux-progress-color`), following the same approach the callout component uses for its colors. The `match()` now sets the variable on the outer element, and the bar reads from it:

```php
$trackClasses = Flux::classes()
    ->add(match ($color) {
        'red'   => '[--flux-progress-color:var(--color-red-600)] dark:[--flux-progress-color:var(--color-red-400)]',
        'green' => '[--flux-progress-color:var(--color-green-600)] dark:[--flux-progress-color:var(--color-green-400)]',
        // ...
    })
    ;

$barClasses = Flux::classes()
    ->add('bg-[var(--flux-progress-color)]')
    ;
```

Static usage is unchanged:

```blade
<flux:progress value="75" color="blue" />
```

Users can now control the color dynamically by setting the CSS variable from Alpine:

```blade
<flux:progress
    x-bind:value="progress"
    x-bind:style="failed && '--flux-progress-color: var(--color-red-600)'"
/>
```

Fixes #2487

### DIFF
--- a/stubs/resources/views/flux/progress.blade.php
+++ b/stubs/resources/views/flux/progress.blade.php
@@ -9,30 +9,31 @@ $trackClasses = Flux::classes()
     ->add('h-1.5 relative w-full overflow-hidden bg-zinc-200 dark:bg-white/10')
     ->add('[print-color-adjust:exact]')
     ->add('rounded-full')
+    ->add(match ($color) {
+        'red'     => '[--flux-progress-color:var(--color-red-600)] dark:[--flux-progress-color:var(--color-red-400)]',
+        'orange'  => '[--flux-progress-color:var(--color-orange-600)] dark:[--flux-progress-color:var(--color-orange-400)]',
+        'amber'   => '[--flux-progress-color:var(--color-amber-600)] dark:[--flux-progress-color:var(--color-amber-400)]',
+        'yellow'  => '[--flux-progress-color:var(--color-yellow-600)] dark:[--flux-progress-color:var(--color-yellow-400)]',
+        'lime'    => '[--flux-progress-color:var(--color-lime-600)] dark:[--flux-progress-color:var(--color-lime-400)]',
+        'green'   => '[--flux-progress-color:var(--color-green-600)] dark:[--flux-progress-color:var(--color-green-400)]',
+        'emerald' => '[--flux-progress-color:var(--color-emerald-600)] dark:[--flux-progress-color:var(--color-emerald-400)]',
+        'teal'    => '[--flux-progress-color:var(--color-teal-600)] dark:[--flux-progress-color:var(--color-teal-400)]',
+        'cyan'    => '[--flux-progress-color:var(--color-cyan-600)] dark:[--flux-progress-color:var(--color-cyan-400)]',
+        'sky'     => '[--flux-progress-color:var(--color-sky-600)] dark:[--flux-progress-color:var(--color-sky-400)]',
+        'blue'    => '[--flux-progress-color:var(--color-blue-600)] dark:[--flux-progress-color:var(--color-blue-400)]',
+        'indigo'  => '[--flux-progress-color:var(--color-indigo-600)] dark:[--flux-progress-color:var(--color-indigo-400)]',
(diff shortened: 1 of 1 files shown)


as247:
Hi @ganyicz why don't make color attribute reactively?
So we can use it like this

```javascript
<flux:progress
    x-bind:value="progress"
    x-bind:color="failed?'red':'blue'"
/>
```
I think this can be done by update flux.js for UIProgress element to track change of color attribute
```javascript
var UIProgress = class extends UIElement {
      ...
    updateVisual() {
      let max2 = this.hasAttribute("max") ? parseFloat(this.getAttribute("max")) : 100;
      let pct = Math.max(0, Math.min(100, this.currentValue / max2 * 100));
      this.style.setProperty("--flux-progress", pct);
      this.style.setProperty("--flux-progress-percentage", pct + "%");
      setAttribute2(this, "aria-valuenow", this.currentValue);
      setAttribute2(this, "aria-valuemax", max2);
      updateColor(this.getAttribute("color"))
    }
  };
```

ganyicz:
Hi @as247,

Unfortunately that's not possible. The color isn't an HTML attribute, it's a Blade prop. 

However you can change the color reactively using the data- attribute as described in the PR

as247:
@ganyicz I found that change color will reset progress bar
before change color: 
```html
style="--flux-progress: 50; --flux-progress-percentage: 50%;"
```
after: 
```html
style="--flux-progress-color: var(--color-red-600)"
```
progress percentage is removed

I have tried using custom html it's working for my custom element:
<x-myicon x-bind:error="isError" x-bind:active="isActive"> error, active also not standard html attribute but MutationObserver is working
```js
new MutationObserver(update)
            .observe($el,{
                attributes:true,
                attributeFilter:['active','error']
})
```
 Then the same way should work for color, we can track the change of color attribute by 
 ```js
 new MutationObserver(update)
            .observe($el,{
                attributes:true,
                attributeFilter:['color']
            })
 ```

Issue #2427, opened 2026-02-22
Modal hides content below the fold with no indication of overflow


schoberg:
### Flux version

2.10.2

### Livewire version

3.7.6

### Tailwind version

4.1.18

### Browser and Operating System

Chrome on macOS

### What is the problem?

When a modal's content is taller than the viewport, it can look visually complete. It's a bad UX pattern and users don't know to scroll depending on how the content hits at the cutoff.  (easiest to visualize in the video below)

The native `<dialog>` element constrains content to the viewport with internal scrolling. But the modal's padding, rounded corners, and shadow all render within that box, so the bottom looks like the natural end.

A related issue: when content grows dynamically (accordion, async load), the centering causes the modal to jump/reposition. 

The common UI patter to solve this is the `<dialog>` becomes a transparent full-viewport scroll container, and the visible modal is an inner panel. The whole modal scrolls as a unit, making overflow obvious and eliminating the jumping issue.

### Code snippets to replicate the problem

Presents with any long content modal or any dynamic content modal (`top` varient fix)

```blade
<flux:modal name="terms" class="md:w-xl">
    <div class="space-y-4">
        <flux:heading size="lg">Terms of Service</flux:heading>
        <flux:text>Please review and accept the following terms.</flux:text>

        <flux:subheading size="lg">1. Introduction</flux:subheading>
        <flux:text>Welcome to our platform. These Terms of Service govern your use of our website and services. By accessing or using our services, you agree to be bound by these terms.</flux:text>

        <flux:subheading size="lg">2. Account Registration</flux:subheading>
        <flux:text>To access certain features, you must register for an account. You agree to provide accurate, current, and complete information during registration.</flux:text>

        <!-- Repeat sections 3-10 so content extends well below the viewport -->
        <!-- The modal will look complete after section 5 or so. Nothing indicates sections 6-10 exist. -->
    </div>
</flux:modal>
```

### Screenshots/ screen recordings of the problem

Here's a 2 min video explaining both issues and showing the fixes: 
https://img.dropinblog.com/FDGMtS65 

Screen recording has 4 short parts:

Problem: tall modal looks complete, content hidden below
Fix: same modal scrolls as a unit, overflowing obvious
Problem: expanding content causes modal to jump causing disorientation
Fix: modal stays pinned, grows downward when using `top` varient

### How do you expect it to work?

The modal should scroll as a visible unit so it's immediately obvious when content extends below the fold.

[...]

PR #2519, opened 2026-03-23
Add scroll="body" prop to modal

ganyicz:
# The scenario

When a modal's content overflows the viewport, there's no visual indication that more content exists below the fold.

# The problem

The rounded corners and padding on the dialog make the bottom edge look like the content ends naturally — even when entire sections and action buttons are hidden below.

# The solution

A new `scroll="body"` prop (following [MUI's convention](https://mui.com/material-ui/react-dialog/#scrolling-long-content)):

```blade
<flux:modal name="terms" scroll="body" class="md:w-xl">
    ...
</flux:modal>
```

When `scroll="body"` is set, the template adds a new branch that wraps the slot in a centering div and a content panel. The `<dialog>` becomes a fullscreen transparent scroll container and the visible card becomes the inner panel:

```blade
<?php if ($overflow): ?>
    <div class="flex min-h-full items-center justify-center p-4 sm:p-6">
        <div {{ $contentAttributes->class($panelClasses) }} data-flux-modal-content>
            {{ $slot }}
        </div>
    </div>
<?php else: ?>
    {{ $slot }}
<?php endif; ?>
```

The dialog is targeted with `data-flux-modal-overflow` for the fullscreen/transparent overrides, and the content panel is targeted with `data-flux-modal-content` for the scale transition and click-outside detection:

```css
[data-flux-modal] > dialog[data-flux-modal-overflow] {
    max-width: none;
    max-height: none;
    width: 100%;
    height: 100%;
    margin: 0;
    overflow-y: auto;
    background: transparent;
    transform: none;
}
```

In `dialogable.js`, click-outside detection uses the content panel's bounding rect when present, falling back to the dialog for existing modals:

```js
let content = this.el.querySelector('[data-flux-modal-content]')
let clickTarget = content || this.el

if (content ? content.contains(e.target) : e.target !== this.el) {
    lastMouseDownEvent = null
    return
}

if (lastMouseDownEvent && clickHappenedOutside(clickTarget, ...) && clickHappenedOutside(clickTarget, ...)) {
    this.cancel()
}
```

Since user attributes like `class` and `style` need to target the visible panel rather than the invisible fullscreen dialog, we split the already-extracted style attributes into content attributes and dialog attributes. For non-overflow modals these are merged back so the public API is unchanged:

```php
[ $contentAttributes, $attributes ] = Flux::splitAttributes($attributes, ['autofocus', 'class', 'style']);
[ $dialogAttributes, $attributes ] = Flux::splitAttributes($attributes, ['wire:close', 'x-on:close', 'wire:cancel', 'x-on:cancel']);

if (! $overflow) {
    $dialogAttributes = $dialogAttributes->merge($contentAttributes->getAttributes());
}
```

Related: livewire/flux-pro — `session/reverent-aurora`

Fixes livewire/flux#2427


ganyicz:
Superseded by new PR with updated screenshots

PR #2520, opened 2026-03-23
Add scroll="body" prop to modal

ganyicz:
# The scenario

When modal content extends beyond the viewport, there's no visual indication that more content exists below the fold.

# The problem

The rounded corners and padding on the dialog make the bottom edge look like the content ends naturally — even when entire sections and action buttons are hidden below.

# The solution

A new `scroll="body"` prop (following [MUI's convention](https://mui.com/material-ui/react-dialog/#scrolling-long-content)):

```blade
<flux:modal name="terms" scroll="body">
    ...
</flux:modal>
```

This allows the modal to overflow and makes the entire viewport scrollable:

https://github.com/user-attachments/assets/15e0eb2a-4779-4e4d-b17f-8727ae28c2fd

Internally, this maps to `$overflow = true` and `dialog[data-flux-modal-overflow]`.

The template wraps the slot in a centering div and a content panel:

```blade
<?php if ($overflow): ?>
    <div class="flex min-h-full items-center justify-center p-4 sm:p-6">
        <div {{ $contentAttributes->class($contentClasses) }} data-flux-modal-content>
            {{ $slot }}
        </div>
    </div>
<?php else: ?>
    {{ $slot }}
<?php endif; ?>
```

Dialog with `data-flux-modal-overflow`  resets sizing and transitions:

```css
[data-flux-modal] > dialog[data-flux-modal-overflow] {
    max-width: none;
    max-height: none;
    width: 100%;
    height: 100%;
    margin: 0;
    overflow-y: auto;
    background: transparent;
    transform: none;
}
```

The content container with newly added `data-flux-modal-content` gets the scale transition:

```css
[data-flux-modal-content] {
    transform: scale(0.95);
    transition: transform 0.075s allow-discrete;
}
```

This data-attribute is also used for click-outside detection:

```js
let contentEl = this.el.querySelector('[data-flux-modal-content]') || this.el

this.el.addEventListener('click', e => {
    if (e.target !== contentEl && contentEl?.contains(e.target)) return
    
    // ...
}
```

Since attributes like `class` and `style` need to target the visible panel, we need an additional split to separate these from other attributes. This is not ideal but not particularly worse than it was.

```php
[ $contentAttributes, $attributes ] = Flux::splitAttributes($attributes, ['autofocus', 'class', 'style']);
[ $dialogAttributes, $attributes ] = Flux::splitAttributes($attributes, ['wire:close', 'x-on:close', 'wire:cancel', 'x-on:cancel']);

if (! $overflow) {
    $dialogAttributes = $dialogAttributes->merge($contentAttributes->getAttributes());
}
```

## Breaking changes

There are no breaking changes. All additions are compatible with previous structure that users may have published.

Fixes livewire/flux#2427

### DIFF
--- a/dist/flux.css
+++ b/dist/flux.css
@@ -76,6 +76,36 @@ Transition dialogs with fade a slight scale effect...
     background-color: rgba(0, 0, 0, 0.10);
 }
 
+/*
+Transition content of dialogs with scroll="body"
+*/
+[data-flux-modal] > dialog[data-flux-modal-overflow] {
+    max-width: none;
+    max-height: none;
+    width: 100%;
+    height: 100%;
+    margin: 0;
+    overflow-y: auto;
+    background: transparent;
+    transform: none;
+}
+
+[data-flux-modal-content] {
+    transform: scale(0.95);
+    transition: transform 0.075s allow-discrete;
+}
+
+dialog[open] [data-flux-modal-content] {
+    transform: scale(1);
+    transition: transform 0.15s allow-discrete;
+}
+
+@starting-style {
+    dialog[open] [data-flux-modal-content] {
+        transform: scale(0.95);
+    }
+}
+
 /**
 Native select styles...
 */


schoberg:
Thanks for shipping this one boys!

PR #2553, opened 2026-04-08
Revert "Encapsulate version checking for view namespace hash"

joshhanley:
Reverts livewire/flux#2551

This PR caused errors, so need to revert it.

Screenshot 2026-04-09 at 08 12 09AM@2x

### DIFF
--- a/src/FluxManager.php
+++ b/src/FluxManager.php
@@ -176,12 +176,11 @@ class FluxManager
 
     public function componentExists($name)
     {
-        return app('view')->exists($this->getFluxViewNamespaceHash() . '::' . $name);
-    }
-
-    public function getFluxViewNamespaceHash()
-    {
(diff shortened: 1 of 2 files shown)


ghabriel25:
@joshhanley Really sorry about that.

Thats really weird. I tested it with my local  project also fresh laravel (with and without blaze) and it works as expected.

joshhanley:
@ghabriel25 all good! The issue was that the `Flux::getFluxViewNamespaceHash();` should have been inside the compiled string in `FluxTagCompiler`, not out of it. But we decided as it's such a small change that we'd revert it instead. The compiler will be quicker with the calculation happening directly rather than going through the `Flux::` facade.

ghabriel25:
I can see the issue now. You're right, it should be at runtime not at compile time

PR #2554, opened 2026-04-09
Update pagination wording to use trans_choice for optional pluralisat…

iainco:
…ion of the word result

# The scenario

This is really picky but something I noticed when working with the table component when paginating and there is only one result in the paginated collection.

# The problem

When using pagination with a table like so:

` <flux:table :paginate="$this->submissions">...</flux:table>`

When the resultset size is 1, the pagination text shows as "Showing 1 to 1 of 1 results", when it should be "Showing 1 to 1 of 1 result".

# The solution

This PR fixes the issue by using the trans_choice helper to choose "result" or "results" correctly.

### DIFF
--- a/stubs/resources/views/flux/pagination.blade.php
+++ b/stubs/resources/views/flux/pagination.blade.php
@@ -72,7 +72,7 @@ $scrollIntoViewJsSnippet = ($scrollTo !== null && $scrollTo !== false)
     <div {{ $attributes->class('@container pt-3 border-t border-zinc-100 dark:border-zinc-700 flex justify-between items-center gap-3') }} data-flux-pagination>
         @if ($paginator->total() > 0)
             <div class="text-zinc-500 dark:text-zinc-400 text-xs font-medium whitespace-nowrap">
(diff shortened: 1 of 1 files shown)


joshhanley (changes requested):
@iainco thanks for the PR!

Not sure if the syntax is correct on this and believe it would currently be a breaking change. I think it should be used as:

```php
trans_choice('results', $paginator->total())
```

That way anyone with "results" translated won't be impacted, but for anyone who wants `trans_choice` to work, would add a translation key like this:

```php
// lang/en/messages.php
'results' => 'result|results',
```
We do try to keep our pagination in line with Laravel's as much as possible, and currently Laravel still has `__('results')`. Might be worth also submitting this change there? 😁https://github.com/laravel/framework/blob/13.x/src/Illuminate/Pagination/resources/views/tailwind.blade.php

calebporzio:
yeah lets just stay inline with laravel's paginator. if it changes there we can change it here. thanks for the PR!

iainco:
Thank you @joshhanley and @calebporzio for your comments, I overlooked how the trans_choice helper works, apologies. Cheers

PR #2556, opened 2026-04-14
Fix flux separator container attributes

### LINKED ISSUE #2555: Text inside `flux:separator` not animated using `wire:transition`
### Flux version

v2.13

### Livewire version

v4.2

### Tailwind version

v4.1

Chrome on Windows

### What is the problem?

When using `wire:transition` attribute on `flux:separator`, the text inside separator doesnt get animated (only the line). This breaks the smooth transition behavior such on validation error login form

```blade
<form class="space-y-4" wire:submit="login">
    <flux:input icon="envelope"
                label="Email"
                placeholder="Enter your e-mail address"
                type="email"
                wire:model="email" />

    <flux:button class="w-full"
                color="green"
                data-test="signin"
                type="submit"
                variant="primary"
                wire:transition>Sign In</flux:button>

    <flux:separator text="or" wire:transition />

    <flux:button class="w-full"
                href="{{ route('google.redirect', absolute: false) }}"
                icon="google"
                wire:transition>
        Sign In with Google
    </flux:button>
</form>
```

https://github.com/user-attachments/assets/53682534-be58-4a0d-828b-c0ec18fefb20

The reason behind this because attribute bag not applied at the container element, it only applied on before and next sibling of the text wrapper. It should be applies on the container as well for case like `wire:text`, `wire:show` etc.

ghabriel25:
Add attribute bag to container element using `container:` prefix

Fixes https://github.com/livewire/flux/issues/2555

### DIFF
--- a/stubs/resources/views/flux/separator.blade.php
+++ b/stubs/resources/views/flux/separator.blade.php
@@ -24,7 +24,8 @@ $classes = Flux::classes('border-0 [print-color-adjust:exact]')
 @endphp
 
 <?php if ($text): ?>
-    <div data-orientation="{{ $orientation }}" {{ $attributes->class('flex items-center w-full') }} role="none" data-flux-separator>
+    <?php $containerAttributes = Flux::attributesAfter('container:', $attributes); ?>
+    <div data-orientation="{{ $orientation }}" {{ $containerAttributes->class('flex items-center w-full') }} role="none" data-flux-separator>
         <div {{ $attributes->class([$classes, 'grow']) }}></div>
 
         <span class="shrink mx-6 font-medium text-sm text-zinc-500 dark:text-zinc-300 whitespace-nowrap">{{ $text }}</span>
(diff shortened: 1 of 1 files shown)


stubs/resources/views/flux/separator.blade.php
joshhanley: I'm concerned about this as we already have `$attributes` forwarding to the separator line divs.

joshhanley (changes requested):
@ghabriel25 thanks for the PR! I don't think we can just add `$attributes` to the parent div here as `$attributes` is already being forwarded to the separator line divs.

We do have a precedent for using a `container`: prefix, the autocomplete and table components both use that, as a work around. So something like this would work:

```blade
<flux:separator container:wire:transition />
```

The way you'd do it is add this to the top of the component

```blade
<?php $containerAttributes = Flux::attributesAfter('container:', $attributes); ?>
```

And change your `$attributes` to `$containerAttributes`.

ghabriel25:
@joshhanley I see, great to know about that! I'll fix that later

joshhanley (approved):

joshhanley:
Thanks!

calebporzio:
mmmm... idk about this one. it's weird that container:class would only work if you had text prop... it's our fault for adding a wrapper conditionally for a prop like that, and I don't want to use conatiner: pattern if at all avoidable.

I'm going to close this one for now. thanks

PR #2564, opened 2026-04-16
Fix Blaze folding `flux:error` as static hidden HTML in inline field templates

### LINKED ISSUE #2560: flux:checkbox validation errors never render (Blaze folds error div as static hidden HTML)
## Bug

Validation errors on `flux:checkbox` (and any other component that uses `flux:with-inline-field`) never render. The `data-flux-error` div is compiled to static hidden HTML and never re-evaluated at runtime.

**Version:** `livewire/flux@v2.13.2`

## Reproduction

```blade
<form wire:submit="save">
    <flux:checkbox label="Accept terms" wire:model="terms" />
    <flux:button type="submit">Submit</flux:button>
</form>
```

Component:

```php
#[Validate('accepted')]
public bool $terms = false;

public function save()
{
    $this->validate();
}
```

Submit without checking the box. Livewire populates the error bag (`errors.terms` is present in the component memo), but the rendered HTML contains:

```html
<div role="alert" ... class="mt-3 text-sm ... hidden" data-flux-error></div>
```

— empty and permanently hidden.

## Root cause

`stubs/resources/views/flux/with-inline-field.blade.php` renders `flux:error` inside an `@blaze(fold: true)` block without wrapping it in `@unblaze`:

```blade
@blaze(fold: true, unsafe: ['name', 'label', 'description'])

@props([
    'name' => \$attributes->whereStartsWith('wire:model')->first(),
    ...
])

...
    <flux:error :\$name />
...
```

At compile time Blaze folds with an empty attribute bag, so the `name` prop default (`whereStartsWith('wire:model')->first()`) resolves to `null`. `flux:error` then renders its empty/hidden branch and Blaze bakes that into the compiled template. At runtime `\$name` is never re-evaluated, so validation errors never display.

`stubs/resources/views/flux/with-field.blade.php` (used by `flux:input`) already handles this correctly by wrapping `flux:error` in `@unblaze(scope: [...])`, which is why input errors work but checkbox errors do not.

## Proposed fix

Apply the same `@unblaze` pattern in `with-inline-field.blade.php`:

```blade
@unblaze(scope: ['name' => \$name])
<flux:error :name=\"\$scope['name']\" />
@endunblaze
```

Happy to send a PR if useful.

## Workaround

Override the stub locally at `resources/views/flux/with-inline-field.blade.php` with the `@unblaze` fix applied. Flux's `anonymousComponentPath` resolves user overrides before the vendor stub.

joshhanley:
# The Scenario

When Blaze is installed, validation errors on `flux:checkbox`, `flux:radio`, and `flux:switch` never render. The error `div` is folded as static hidden HTML at compile time and never re-evaluated at runtime.

```php
<?php

use Livewire\Attributes\Validate;
use Livewire\Component;

new class extends Component {
    #[Validate('accepted')]
    public bool $terms = false;

    public function save()
    {
        $this->validate();
    }
}; ?>

<form wire:submit="save">
    <flux:checkbox label="Accept terms" wire:model="terms" />
    <flux:button type="submit">Submit</flux:button>
</form>
```

# The Problem

`with-inline-field.blade.php` and `with-reversed-inline-field.blade.php` render `flux:error` directly inside a `@blaze(fold: true)` block without wrapping it in `@unblaze`. `with-field.blade.php` already handles this correctly.

# The Solution

Wrap `flux:error` in `@unblaze` in both `with-inline-field.blade.php` and `with-reversed-inline-field.blade.php`, matching the existing pattern in `with-field.blade.php`.

Fixes #2560

### DIFF
--- a/stubs/resources/views/flux/with-inline-field.blade.php
+++ b/stubs/resources/views/flux/with-inline-field.blade.php
@@ -28,7 +28,9 @@ extract(Flux::forwardedattributes($attributes, [
             <flux:description>{{ $description }}</flux:description>
         <?php endif; ?>
 
-        <flux:error :$name />
+        @unblaze(scope: ['name' => $name])
+        <flux:error :name="$scope['name']" />
+        @endunblaze
     </flux:field>
 <?php else: ?>
     {{ $slot }}
--- a/stubs/resources/views/flux/with-inline-field.blade.php
+++ b/stubs/resources/views/flux/with-inline-field.blade.php
@@ -29,7 +29,7 @@ extract(Flux::forwardedattributes($attributes, [
         <?php endif; ?>
 
         @unblaze(scope: ['name' => $name])
-        <flux:error :name="$scope['name']" />
+            <flux:error :name="$scope['name']" />
         @endunblaze
     </flux:field>
 <?php else: ?>
--- a/stubs/resources/views/flux/with-reversed-inline-field.blade.php
+++ b/stubs/resources/views/flux/with-reversed-inline-field.blade.php
@@ -29,7 +29,9 @@ extract(Flux::forwardedattributes($attributes, [
 
         {{ $slot }}
 
-        <flux:error :$name />
+        @unblaze(scope: ['name' => $name])
+        <flux:error :name="$scope['name']" />
+        @endunblaze
     </flux:field>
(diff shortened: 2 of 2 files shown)


calebporzio:
Thanks Josh!

PR #2581, opened 2026-04-20
Support `wire:model` on expandable `flux:sidebar.group` and `flux:navlist.group`

### LINKED ISSUE #2580: Cannot targeting `flux:sidebar.group` button for persisting the expanded state
### Flux version

v2.13

### Livewire version

v4.2

### Tailwind version

v4.2

Chrome on Window

### What is the problem?

I'm trying to persist the expanded state of sidebar group but there is no option for that. Then I made it as Livewire component with `#[Session]` so I can toggle the `expanded` attribute and persisted.

However, there is no way to targeting the button inside sidebar group. Here's what I have tried
- Use `wire:click="$toggle('expanded')"` on sidebar group
- It works but it also triggered from child element click event
- Try to add `.self` modifier, there was no error but nothing happened 
(No network request, `expanded` not changed)

```php
<?php

use Livewire\Component;

new class extends Component
{
    #[Session]
    public bool $expanded = true;
}
```
```blade
<flux:sidebar.group :expanded="$expanded"
                    expandable
                    heading="My Items"

                    {{-- also get triggered from child click event --}}
                    {{-- using .self modifier doesnt work --}}
                    wire:click="$toggle('expanded')"> 
    ...
</flux:sidebar.group>
```

### Workaround
```blade
<flux:sidebar.group :expanded="$expanded"
                    expandable
                    heading="My Items"
                    x-data="{ button: $el.querySelector('button') }"
                    x-init="button.setAttribute('wire:click', `$toggle('expanded')`)">
    ...
</flux:sidebar.group>
```

N/A

I expect `wire:click` attribute on sidebar group getting passed down to the button that trigger the sidebar group expanded or not

joshhanley:
# The Scenario

A developer wants to persist whether an expandable `flux:sidebar.group` or `flux:navlist.group` is open by binding it to a Livewire property. Adding `wire:click` to the group header doesn't work cleanly: clicks on child items bubble up to the wrapper and re-trigger the toggle, and `.self` doesn't help because the actual click target is an inner `<button>`, not the wrapper.

```php
<?php

use Livewire\Component;
use Livewire\Attributes\Session;

new class extends Component
{
    #[Session]
    public bool $expanded = true;
}; ?>

<flux:sidebar.group :expanded="\$expanded"
                    expandable
                    heading="My Items"
                    wire:click="\$toggle('expanded')">
    <flux:sidebar.item href="/home">Home</flux:sidebar.item>
    <flux:sidebar.item href="/settings">Settings</flux:sidebar.item>
</flux:sidebar.group>
```

# The Problem

Both `stubs/resources/views/flux/sidebar/group.blade.php` and `stubs/resources/views/flux/navlist/group.blade.php` spread `\$attributes` directly onto the outer `<ui-disclosure>` wrapper. `wire:model` would therefore land on `<ui-disclosure>`, but without the `.self` modifier it would also pick up bubbled `input`/`change` events from nested form controls inside the group.

The sibling component `flux:accordion.item` already handles this exact case by auto-applying the `.self` modifier to `wire:model` and setting up Alpine `x-data` / `x-model.self` for clean two-way binding with the disclosure's `value`. `flux:sidebar.group` and `flux:navlist.group` were missing this.

# The Solution

Apply the same pattern from `flux:accordion.item` to both `flux:sidebar.group` and `flux:navlist.group` when `expandable`:

- Auto-add `.self` to any `wire:model` directive so it only listens to events fired directly on `<ui-disclosure>`.
- Initialise Alpine `x-data="{ open: \$wire.<prop> }"` and bind via `x-model.self="open"` so the disclosure value stays in sync with the Livewire property in both directions.

After this change, the original use case becomes:

```blade
<flux:sidebar.group expandable heading="My Items" wire:model.live="expanded">
    <flux:sidebar.item href="/home">Home</flux:sidebar.item>
    <flux:sidebar.item href="/settings">Settings</flux:sidebar.item>
</flux:sidebar.group>
```

Fixes #2580

### DIFF
--- a/stubs/resources/views/flux/navlist/group.blade.php
+++ b/stubs/resources/views/flux/navlist/group.blade.php
@@ -6,8 +6,24 @@
     'heading' => null,
 ])
 
+@php
+$wireModel = $attributes->wire('model');
+
+// Support adding the .self modifier to the wire:model directive...
+if ($expandable && $wireModel && $wireModel->directive && ! $wireModel->hasModifier('self')) {
+    unset($attributes[$wireModel->directive]);
+
+    $wireModel->directive .= '.self';
+
+    $attributes = $attributes->merge([$wireModel->directive => $wireModel->value]);
+}
+
+// Support binding the state to a Livewire property
+$state = $wireModel?->value ? '$wire.' . $wireModel->value : ($expanded ? 'true' : 'false');
+@endphp
+
 <?php if ($expandable && $heading): ?>
-    <ui-disclosure {{ $attributes->class('group/disclosure') }} @if ($expanded === true) open @endif data-flux-navlist-group>
+    <ui-disclosure {{ $attributes->class('group/disclosure') }} x-data="{ open: {{ $state }} }" x-model.self="open" @if ($expanded === true) open @endif data-flux-navlist-group>
         <button type="button" class="w-full h-10 lg:h-8 flex items-center group/disclosure-button mb-[2px] rounded-lg hover:bg-zinc-800/5 dark:hover:bg-white/[7%] text-zinc-500 hover:text-zinc-800 dark:text-white/80 dark:hover:text-white">
             <div class="ps-3 pe-4">
                 <flux:icon.chevron-down class="size-3! hidden group-data-open/disclosure-button:block" />
--- a/stubs/resources/views/flux/sidebar/group.blade.php
+++ b/stubs/resources/views/flux/sidebar/group.blade.php
@@ -12,9 +12,25 @@
     'icon' => null,
 ])
 
+@php
+// Support adding the .self modifier to the wire:model directive...
+if ($expandable && ($wireModel = $attributes->wire('model')) && $wireModel->directive && ! $wireModel->hasModifier('self')) {
+    unset($attributes[$wireModel->directive]);
+
+    $wireModel->directive .= '.self';
+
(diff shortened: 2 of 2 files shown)


calebporzio:
Hmmm idk about this.

A) I don't like the forwarding to `x-data` pattern. I understand we have a precedent for it, so it's not a total deal breaker but not ideal
B) What we really want is persistance here and we can achieve that in less verbose ways than a wire:model.live and a session attribute property - that's a lot of ceremony. (and network requests)

My instinct is to either:
A) Ignore this and close till others request this
B) Explore piggybacking on the existing `persist` prop for the sidebar component itself. Like we can just also store the expanded state of a group in localStorage like we do expanded state of the sidebar itself.

The only tricky part about option "B" is if someone wanted to opt-out of it, they would also be opting out of localStorage expanded state for the entire sidebar component as well.

We could support setting `persist` on both the sidebar and sidebar groups but that could feel like too much. especially for lots of groups.

There's probably also technical tricky bits like how to key a sidebar group in local storage deterministically if there is no name provided.

---

The most important thing to do here before anything else is to look at lots of common and popular and well built websites that have sidebar groups and observe how they behave.

If many don't persist expanded state, we are left still deciding what to do. If they all support it, the decision is easy: we support it. Then we just have to figure out the best way.

PR #2583, opened 2026-04-21
Forward `icon:variant` to navbar item trailing icon

ghabriel25:
This PR modify `flux:navbar.item` and forward `icon:variant` to its trailing icon which should be consistence with
- `flux:navlist.item`
- `flux:navmenu.item`

As comparison
### `flux:navbar.item`
```blade

<?php if (is_string($iconTrailing) && $iconTrailing !== ''): ?>
    <flux:icon :icon="$iconTrailing" variant="micro" class="size-4 ms-1" />
    {{-- Update: <flux:icon :icon="$iconTrailing" :variant="$iconVariant" class="{!! $iconClasses !}}" /> --}}
<?php elseif ($iconTrailing): ?>
    {{ $iconTrailing }}
<?php endif; ?>
```
### `flux:navlist.item`
```blade
@php
    ...
    // Size-up icons in square/icon-only buttons...
    $iconClasses = Flux::classes($square ? 'size-5!' : 'size-4!');
    ...
@endphp

<?php if (is_string($iconTrailing) && $iconTrailing !== ''): ?>
    <flux:icon :icon="$iconTrailing" :variant="$iconVariant" class="size-4!" />
<?php elseif ($iconTrailing): ?>
    {{ $iconTrailing }}
<?php endif; ?>
```
### `flux:navmenu.item`
```blade
@php
    ...
    $iconClasses = Flux::classes()
        ->add('me-2')
        // When using the outline icon variant, we need to size it down to match the default icon sizes...
        ->add($iconVariant === 'outline' ? 'size-5' : null)
        ;

    $trailingIconClasses = Flux::classes()
        ->add('ms-auto')
        // When using the outline icon variant, we need to size it down to match the default icon sizes...
        ->add($iconVariant === 'outline' ? 'size-5' : null)
        ;
    ...
@endphp

<?php if (is_string($iconTrailing) && $iconTrailing !== ''): ?>
    <flux:icon :icon="$iconTrailing" :variant="$iconVariant" :class="$trailingIconClasses" />
<?php elseif ($iconTrailing): ?>
    {{ $iconTrailing }}
<?php endif; ?>
```

```blade
// icon:variant not forwarded
<flux:navbar.item icon:variant="outline" icon:trailing="bell-alert">Notifications</flux:navbar.item>
```

| Before | After |
| - | - |
| image | image |

### DIFF
--- a/stubs/resources/views/flux/navbar/item.blade.php
+++ b/stubs/resources/views/flux/navbar/item.blade.php
@@ -22,7 +22,7 @@
 $square ??= $slot->isEmpty();
 
 // Size-up icons in square/icon-only buttons...
-$iconClasses = Flux::classes($square ? 'size-6' : 'size-5');
+$iconClasses = Flux::classes($square ? 'size-6!' : 'size-5!');
 
 $classes = Flux::classes()
     ->add('px-3 h-8 flex items-center rounded-lg')
@@ -72,7 +72,7 @@ $classes = Flux::classes()
     <?php endif; ?>
 
     <?php if (is_string($iconTrailing) && $iconTrailing !== ''): ?>
-        <flux:icon :icon="$iconTrailing" :variant="$iconVariant" class="size-4! ms-1" />
+        <flux:icon :icon="$iconTrailing" :variant="$iconVariant" class="size-5! ms-1" />
     <?php elseif ($iconTrailing): ?>
         {{ $iconTrailing }}
     <?php endif; ?>
--- a/stubs/resources/views/flux/navbar/item.blade.php
+++ b/stubs/resources/views/flux/navbar/item.blade.php
@@ -72,7 +72,7 @@ $classes = Flux::classes()
     <?php endif; ?>
 
     <?php if (is_string($iconTrailing) && $iconTrailing !== ''): ?>
-        <flux:icon :$icon :variant="$iconVariant" class="{!! $iconClasses !!}" />
+        <flux:icon :icon="$iconTrailing" :variant="$iconVariant" class="{!! $iconClasses !!}" />
     <?php elseif ($iconTrailing): ?>
         {{ $iconTrailing }}
     <?php endif; ?>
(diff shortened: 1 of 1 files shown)


stubs/resources/views/flux/navbar/item.blade.php
```diff
@@ -22,7 +22,7 @@
 $square ??= $slot->isEmpty();
 
 // Size-up icons in square/icon-only buttons...
-$iconClasses = Flux::classes($square ? 'size-6' : 'size-5');
+$iconClasses = Flux::classes($square ? 'size-6!' : 'size-5!');
```
joshhanley: We shouldn't be using `!` in Flux code.

stubs/resources/views/flux/navbar/item.blade.php
```diff
@@ -72,7 +72,7 @@
     <?php endif; ?>
 
     <?php if (is_string($iconTrailing) && $iconTrailing !== ''): ?>
-        <flux:icon :icon="$iconTrailing" variant="micro" class="size-4 ms-1" />
+        <flux:icon :icon="$iconTrailing" :variant="$iconVariant" class="size-5! ms-1" />
```
joshhanley: If the size was previously 4, we can't bump it up to 5 now. If anything, the trailing icon code should match the leading icon code `<flux:icon :$icon :variant="$iconVariant" class="{!! $iconClasses !!}" />`

joshhanley (changes requested):
@ghabriel25 thanks for the PR! We can't use `!` in Flux code as it means other people can't override it themselves using `!`. I think the trailing icon being different here was intentional, as typically the trailing icon is a down chevron, as documented here https://fluxui.dev/components/navbar#dropdown-navigation

But, I can see an argument to make the trailing icon match the leading icon, in which case you could basically copy the leading icon line and update it to suit.

ghabriel25:
Thanks @joshhanley for the review. I have update it like you said and match trailing icon with leading icon.

joshhanley (approved):
@ghabriel25 thanks!

ghabriel25:
Hi, @joshhanley 

Looking back again at the stubs. I notice that `flux:button` uses `icon-trailing:variant` and I kind of agree with that approach where component leading icon and trailing icon variant should be differentiated if there is `icon-trailing:variant` passed as attribute. If not, just use `icon:variant`.

However, I'm personally okay with the current layout. I just think it could be more flexible about that

what do you think?

PR #2585, opened 2026-04-22
Support `iconDot` for trailing icon in `flux:navbar.item` and `flux:navlist.item`

ghabriel25:
# The scenario

There should be a case when user needs trailing icon with `iconDot` for example in mobile screen

**Note:** There is a difference between trailing icon in `flux:navbar.item` and `flux:navlist.item` (See https://github.com/livewire/flux/pull/2583)

### `flux:navlist.item`
image

### `flux:navbar.item`
image

# The problem

Currently, `iconDot` only for leading icon in `flux:navbar.item` and `flux:navlist.item` but not for trailing icon.

# The solution

Add support `iconDot` prop for trailing icon match with leading icon. There are no change on how navbar and navlist displayed.

# Files changed
- `flux/navbar/item.blade.php`
- `flux/navlist/item.blade.php`


joshhanley (changes requested):
@ghabriel25 thanks for the PR! 

My concern for this is that it's a common pattern to have icons on the left, because the right is used for chevrons, such as for a dropdown menu. But now it is adding icon dots to both (see below).

Image

Maybe it could be `iconTrailingDot` instead?

ghabriel25:
Thanks for the review @joshhanley 

You're right about that. I have update it and add `iconTrailingDot` as a prop

joshhanley (approved):

calebporzio:
I don't love the tripple-compound word: `iconTrailingDot` here.

Can we just use `iconDot` as a signal for either icon? Like when would a user put two icons on a navbar item anyways? 

I feel like we should just make it so that you say `icon-trailing="foo" icon-dot` and it works?

that's my vote.

ghabriel25:
Then I'll change it to initial commit.

@calebporzio Theres a nit about `iconDot` which is user can't customized the color. Is there a plan to make icon dot as dedicated component. It seems this pattern used by some components.

For me, the idea of `iconDot` was to tell the user something is new (or unread etc). So, it should catch the attention. For example on collapsed sidebar, `badge` can be replaced with `iconDot`

joshhanley:
> @ghabriel25 thanks for the PR!
> 
> My concern for this is that it's a common pattern to have icons on the left, because the right is used for chevrons, such as for a dropdown menu. But now it is adding icon dots to both (see below).
> 
> Image
> Maybe it could be `iconTrailingDot` instead?

@calebporzio I agree with the triple word isn't great. But I would say it's very common if you have dropdown menu and want icons on all of your navbar items, as the down chevron is a trailing icon (see this quoted comment). The screenshot here is from when I pulled this PR and tested it and we ended up with dots on both icons, which seems bad to me.

ghabriel25:
@joshhanley @calebporzio How about ignore icon-dot on trailing icon if leading icon exists? Something like
```php
if ($iconDot && ! $icon)
```
However, this only introduce new inconsistency on codebase

calebporzio:
thanks!

PR #2588, opened 2026-04-23
Add empty string checking on `flux:input` leading and trailing icon

ghabriel25:
I notice that almost all component that using leading and trailing icon always checking if its empty string or not. However, `flux:input` component doesnt align with this.

This PR adds empty checking for this component which suppose aligned with other components that using icon.

### Files changed
- `stubs/resources/views/flux/input/index.blade.php`


stubs/resources/views/flux/badge/index.blade.php
ghabriel25, line 1: I failed to notice that this component already has checking.

joshhanley (approved):

PR #2592, opened 2026-04-24
Support `flux:badge` with tooltip

ghabriel25:
`flux:badge` can be set as a `<div>` or a `<button>` however it doesnt support tooltip attribute (yet) meanwhile `flux:button` which can be set as a `<div>`, `<a>` or `<button>` already support tooltip attribute.

This PR wrap `flux:badge` with `flux:with-tooltip` and pass the attributes just like `flux:button`

### Before
https://github.com/user-attachments/assets/a3386e56-c5d9-408e-b7ca-b2d03310d2d7

### After
https://github.com/user-attachments/assets/079ea20d-8faf-453d-9713-346416eba50a

File changed
- `flux/badge/index.blade.php`

### DIFF
--- a/stubs/resources/views/flux/badge/index.blade.php
+++ b/stubs/resources/views/flux/badge/index.blade.php
@@ -86,22 +86,24 @@ $classes = Flux::classes()
     });
 @endphp
 
-<flux:button-or-div :attributes="$attributes->class($classes)" data-flux-badge>
-    <?php if (is_string($icon) && $icon !== ''): ?>
-        <flux:icon :$icon :variant="$iconVariant" :class="$iconClasses" data-flux-badge-icon />
-    <?php else: ?>
-        {{ $icon }}
-    <?php endif; ?>
+<flux:with-tooltip :$attributes>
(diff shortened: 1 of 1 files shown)


calebporzio:
hmmm i hesitate to merge this but it does make sense. can you make sure we aren't missing blaze "unsafe" attributes or something that need to be added to components that are wrapped in "with-tooltip"

in general i like to avoid wrapping components like that, and it might not be worth it if this isn't a high demand feature. idk.

ghabriel25:
Yeah, that’s a fair concern. I took a look specifically at Blaze ‘unsafe’ attributes and didn’t find any cases where wrapping with with-tooltip would strip or block them, since the wrapper just passes props through.

I also get the hesitation around wrapping components—I'm not a huge fan of that pattern either. My thinking was mainly to keep tooltip behavior consistent without duplicating logic, but I agree it might be overkill if this isn’t something we expect to use much.

calebporzio:
yeah ok, let's close for now. But i agree re: consistency. So if there is other demand we can consider adding it. thanks!

PR #2596, opened 2026-04-28
Replace blade directive with php syntax

ghabriel25:
This PR consists of
- Replace the blade directive used in component html with php syntax mainly for consistency with other components. 
- Inline blade directive remains untouched.
- Pagination component remains untouched (I get overwhelmed by its directives 🤕 )

### File changed
- `stubs/resources/views/flux/breadcrumbs/item.blade.php`
- `stubs/resources/views/flux/checkbox/variants/buttons.blade.php`
- `stubs/resources/views/flux/radio/variants/buttons.blade.php`
- `stubs/resources/views/flux/sidebar/index.blade.php`
- `stubs/resources/views/flux/table/sortable.blade.php`
- `stubs/resources/views/flux/header.blade.php`
- `stubs/resources/views/flux/switch.blade.php`

### DIFF
--- a/stubs/resources/views/flux/breadcrumbs/item.blade.php
+++ b/stubs/resources/views/flux/breadcrumbs/item.blade.php
@@ -56,7 +56,7 @@ $iconClasses = Flux::classes()
         </div>
     <?php endif; ?>
 
-    <?php if ($separator === null): ?>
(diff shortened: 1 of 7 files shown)


calebporzio:
great, thanks!

PR #2626, opened 2026-05-20
Fix Blaze safe list in select option

### LINKED ISSUE #2613: flux:select.option ignores dynamic :indicator from parent due to safe: ['indicator'] bypassing Blaze v1.0.12 dynamic-aware abort
## Versions

- `livewire/flux` ^2 (Pro)
- `livewire/blaze` v1.0.12
- `livewire/livewire` v4
- Laravel 12, PHP 8.4

## Summary

When `<flux:select>` receives `indicator` as a **dynamic** value (e.g. `:indicator="$multiple ? 'checkbox' : null"`), the checkbox indicator is **not** rendered inside `<flux:select.option>`. A static value (`indicator="checkbox"`) works as expected.

This use case is exactly what livewire/blaze#172 ("Abort fold when aware prop is dynamic") was supposed to cover — but Flux's own `safe: ['indicator']` declaration neutralises that abort. So the bug Blaze fixed in v1.0.12 still reproduces specifically for `flux:select.option`.

Related earlier issue I filed: livewire/blaze#171.

## Reproduce

Anonymous Blade component that forwards `$multiple` into a dynamic `:indicator`:

```blade
{{-- resources/views/components/filters/dictionary-select.blade.php --}}
@props([
    'fieldLabel',
    'items',
    'multiple' => false,
])

<flux:select
    variant="listbox"
    :label="__($fieldLabel)"
    :multiple="$multiple"
    :indicator="$multiple ? 'checkbox' : null"
>
    @foreach($items as $item)
        <flux:select.option :value="$item->id">{{ $item->name }}</flux:select.option>
    @endforeach
</flux:select>
```

Usage:

```blade
<x-filters.dictionary-select fieldLabel="Districts" :items="$districts" :multiple="true" />
```

**Expected:** each option renders a checkbox indicator (since parent is `multiple` and `indicator="checkbox"` resolves at runtime).
**Actual:** no checkbox indicator. Options render like a default listbox.

Replacing the dynamic prop with a static one (`indicator="checkbox"` on the parent `<flux:select>`) fixes it. Splitting the component into two `@if($multiple) … @else …` branches and putting `indicator="checkbox"` **directly on each `<flux:select.option>`** also fixes it (current workaround on our side).

## Root cause

`flux:select.option/index.blade.php` declares:

```blade
@blaze(fold: true, safe: ['filterable', 'indicator', 'loading'])
@aware([ 'variant', 'indicator' ])
```

Blaze v1.0.12 added a guard in `Folder::isSafeToFold()` that collects `@aware` props from the parent's attributes and forces them into `$dynamicAttributes` so the fold can be aborted when they are dynamic:

```php
foreach ($source->directives->aware() as $prop) {
    if (! isset($node->attributes[$prop])
        && isset($node->parentsAttributes[$prop])
        && ! $node->parentsAttributes[$prop]->isStaticValue()
    ) {
        $dynamicAttributes[$prop] = $node->parentsAttributes[$prop];
    }
}

// ...
$unsafe = array_diff(array_merge($props, $aware, $unsafe), $safe);

foreach ($dynamicAttributes as $attribute) {
    if (in_array($attribute->propName, $unsafe)) {
        return false; // abort fold
    }
}
```

For `flux:select.option` with the repro above:

- `$dynamicAttributes['indicator']` — added by the new guard ✅
- `$aware = ['variant', 'indicator']`
- `$safe = ['filterable', 'indicator', 'loading']`
- After `array_diff(... , $safe)`: `indicator` is **removed** from `$unsafe`
- Final check: `indicator` is not in `$unsafe` → fold is **not** aborted

The fold proceeds, the dynamic `:indicator` placeholder gets lost across the `flux:delegate-component` boundary into `select/option/variants/custom.blade.php`, and `@aware([ 'indicator' ])` there falls back to `null`.

## Why `safe: ['indicator']` is too optimistic

The `safe` list was presumably added to allow folding when `indicator` is passed statically (the common case). But it also overrides the new Blaze-level abort for the legitimate dynamic-`@aware` case. The decision "is folding safe?" should be different for:

- attribute declared on the component itself (current `safe` semantics — fine),
- attribute resolved via `@aware` from a dynamic parent attribute (should ignore `safe` and abort).

## Suggested fix (one of)

1. **Drop `'indicator'` from `safe` in `flux:select/option/index.blade.php`.** Static-`indicator` cases will fall back to the normal fold-safe path; dynamic ones will correctly abort.
2. **Or upstream in Blaze:** in `isSafeToFold`, treat attributes that came from the `@aware`-parent-dynamic branch as always unsafe regardless of the `safe` list (apply the abort *before* `safe` subtraction). Arguably the more robust fix, since other Flux components likely have the same shape.

Happy to send a PR against either repo if you have a preference on direction.

## Workaround in our project

Splitting the anonymous component into two static branches and inlining `indicator="checkbox"` on each `<flux:select.option>` in the multiple branch. Works without touching Flux internals, but loses the elegance of a single dynamic `:indicator`.

ganyicz:
# The scenario

When `indicator` is passed in as a dynamic property, the select always renders the default indicator.

```blade
<flux:select
    variant="listbox"
    :indicator="$multiple ? 'checkbox' : null"
>
    ...
</flux:select>
```

SCR-20260520-tslb

# The problem

The problem is that in the above example, the select folds because `indicator` is currently marked as `safe`. However `indicator` is not a pass-through prop, it changes which indicator component is rendered inside the option.

This was introduced by an oversight in #2496, where we marked props of `flux::delegate-component` targets as `unsafe` in their parent component (because Blaze cannot "see through" the delegate component).

What makes it even clearer that this was an oversight, is that [this was the only component in the entire diff](https://github.com/livewire/flux/commit/d0202a2f05953b9b3d404bcbadadf050e426b381#diff-b6a765c7e42ee17d4efe2d1b7fb2e12436d0e57d0b97641ef8fdc6d36f7a3c1a) that added `safe` props. In all the other components we marked props as `unsafe`, which was the main point of the PR.

# The solution

Change `safe` to `unsafe` + add comment for consistency with other components.

Fixes livewire/flux#2613

### DIFF
--- a/stubs/resources/views/flux/select/option/index.blade.php
+++ b/stubs/resources/views/flux/select/option/index.blade.php
@@ -1,4 +1,5 @@
-@blaze(fold: true, safe: [
+@blaze(fold: true, unsafe: [
+    // variant props
     'filterable', 'indicator', 'loading',
 ])
 


joshhanley:
Ah, thanks for that! I never would have picked up that it was wrong 😅

Issue #2627, opened 2026-05-21
Textarea component breaks CSP by using style attribute


troccoli-sgr:
### Flux version

v2.14.1

### Livewire version

v4.3.0

### Tailwind version

v4.2.2

### Browser and Operating System

Chrome on MacOS

### What is the problem?

This follows discussion #2624.

Basically, when you use a textarea component you can choose how to resize it. This is then translated into using a `style` attribute, e.g. `style="resize: none; "`, rather than using a Tailwind classes, e.g. `resize-none`. In fact, the component always uses the `resize-y` classes, so the logic to use a Tailwind class should already be there.

The component works correctly, and it is resized correctly. But if you enable CSP, then you get an error in the console

```
Applying inline style violates the following Content Security Policy directive 'style-src 'self' 'nonce-9uZmpoBEpEoJcW4PJ1d3dHaBSvN3sjS9' fonts.googleapis.com fonts.bunny.net'. Either the 'unsafe-inline' keyword, a hash ('sha256-taOiJvFwCIELnoVAMuNfu+uAuWSkv20pPWS0piJHcO4='), or a nonce ('nonce-...') is required to enable inline execution. Note that hashes do not apply to event handlers, style attributes and javascript: navigations unless the 'unsafe-hashes' keyword is present. The action has been blocked.
```

### Code snippets to replicate the problem

```blade
<flux:textarea resize="none"/>
```

### Screenshots/ screen recordings of the problem

The HTML produced is

```html
<textarea 
    class="block p-3 w-full shadow-xs disabled:shadow-none border rounded-lg bg-white dark:bg-white/10 dark:disabled:bg-white/[7%] resize-y text-base sm:text-sm text-zinc-700 disabled:text-zinc-500 placeholder-zinc-400 disabled:placeholder-zinc-400/70 dark:text-zinc-300 dark:disabled:text-zinc-400 dark:placeholder-zinc-400 dark:disabled:placeholder-zinc-500 border-zinc-200 border-b-zinc-300/80 dark:border-white/10 data-invalid:shadow-none data-invalid:border-red-500 dark:data-invalid:border-red-500" 
    rows="4" 
    style="resize: none; " 
    data-flux-control="" 
    data-flux-textarea="" 
    aria-describedby="lofi-description-5338a4fa6e8fa8"
></textarea>
```

<img width="902" height="105" alt="Image" src="https://github.com/user-attachments/assets/817f0fe4-1fc4-4f1e-a92b-a7146680f051" />

[...]

PR #2629, opened 2026-05-21
Replace `flux:textarea` inline style with tailwindcss class

ghabriel25:
# The scenario

Although I can't seem to reproduce this issue https://github.com/livewire/flux/issues/2627, I think this might be better to use tailwind class on `flux:textarea` since this style not exposed to Alpine like `flux:progress` in https://github.com/livewire/flux/pull/2512.

We can revert it if this style need to be set dynamically using Alpine. I just never see the need of it, at least for me.

# The problem

`flux:textarea` current template using inline style for resize type and field-sizing.

```blade
<textarea
    ...
    style="{{ $resizeStyle }}; {{ $rows === 'auto' ? 'field-sizing: content' : '' }}"
    ... 
    >{{ $slot }}</textarea>
```

# The solution

Change inline style with tailwindcss class which can cover it all.

```php
$classes = Flux::classes()
    ...
    ->add($rows === 'auto' ? 'field-sizing-content' : '')
    ->add($resize ? match ($resize) {
        default => 'resize', // to prevent unhandled match type when user only provide the attribute without value
        'none' => 'resize-none',
        'both' => 'resize',
        'horizontal' => 'resize-x',
        'vertical' => 'resize-y',
    } : 'resize-none')
    ...
```

# File Changes
- `stubs/resources/views/flux/textarea.blade.php`

### DIFF
--- a/stubs/resources/views/flux/textarea.blade.php
+++ b/stubs/resources/views/flux/textarea.blade.php
@@ -18,25 +18,24 @@ $classes = Flux::classes()
     ->add('block p-3 w-full')
     ->add('shadow-xs disabled:shadow-none border rounded-lg')
     ->add('bg-white dark:bg-white/10 dark:disabled:bg-white/[7%]')
-    ->add($resize ? 'resize-y' : 'resize-none')
+    ->add($rows === 'auto' ? 'field-sizing-content' : '')
+    ->add($resize ? match ($resize) {
+        default => 'resize' // to prevent unhandled match type when user only provide the attribute without value
+        'none' => 'resize-none',
+        'both' => 'resize',
+        'horizontal' => 'resize-x',
+        'vertical' => 'resize-y',
+    } : 'resize-none')
     ->add('text-base sm:text-sm text-zinc-700 disabled:text-zinc-500 placeholder-zinc-400 disabled:placeholder-zinc-400/70 dark:text-zinc-300 dark:disabled:text-zinc-400 dark:placeholder-zinc-400 dark:disabled:placeholder-zinc-500')
     ->add('border-zinc-200 border-b-zinc-300/80 dark:border-white/10')
     ->add('data-invalid:shadow-none data-invalid:border-red-500 dark:data-invalid:border-red-500')
     ;
-
-$resizeStyle = match ($resize) {
-    'none' => 'resize: none',
-    'both' => 'resize: both',
-    'horizontal' => 'resize: horizontal',
-    'vertical' => 'resize: vertical',
-};
 @endphp
 
 <flux:with-field :$attributes>
     <textarea
         {{ $attributes->class($classes) }}
         rows="{{ $rows }}"
-        style="{{ $resizeStyle }}; {{ $rows === 'auto' ? 'field-sizing: content' : '' }}"
         @isset ($name) name="{{ $name }}" @endisset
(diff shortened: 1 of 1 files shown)


stubs/resources/views/flux/textarea.blade.php
```diff
@@ -18,25 +18,24 @@
...
     ->add('bg-white dark:bg-white/10 dark:disabled:bg-white/[7%]')
-    ->add($resize ? 'resize-y' : 'resize-none')
+    ->add($rows === 'auto' ? 'field-sizing-content' : '')
+    ->add($resize ? match ($resize) {
+        default => 'resize', // to prevent unhandled match type when user only provide the attribute without value
+        'none' => 'resize-none',
+        'both' => 'resize',
+        'horizontal' => 'resize-x',
+        'vertical' => 'resize-y',
+    } : 'resize-none')
```
troccoli-sgr, line 27: I think this could be improved.

If `$resize` is not defined this would use `resize-none` but the original default was `resize-y`. 

To make sure that `$resize` is always defined, and defaults to `vertical`, you can use
```php
$resize ??= 'vertical';
```
just before setting the classes, and maybe remove the default in the `props`.

Then this block can be simplified as
```suggestion
    ->add(match ($resize) {
        'none' => 'resize-none',
        'both' => 'resize',
        'horizontal' => 'resize-x',
        'vertical' => 'resize-y',
        default => 'resize-y'
    })
```

what do you thnk?

calebporzio (commented):
Thanks @ghabriel25! Since removing an inline style can be load-bearing, I verified this one behaviorally (fresh Laravel app, Flux as a path repo, computed styles compared on `main` vs this branch):

- `resize` default / `none` / `horizontal` / `both`: computed `resize` values identical in every mode.
- `rows="auto"`: computed `field-sizing: content` intact, and the auto-grow behavior is unchanged (typed 8 lines into the textarea: grew 46px -> 186px on both branches).
- No JS dependency: the only `textarea` reference in Flux's JS is a focusable-elements selector, so nothing reads or writes these inline styles.
- Bonus fix worth calling out: on `main`, `<flux:textarea resize />` (bare attribute) throws an `UnhandledMatchError` because the old `match` has no default arm. This branch renders it fine.

On @troccoli-sgr's note: `$resize` always has a value here — the prop defaults to `'vertical'` — so the default path still produces `resize-y`/`resize: vertical` (confirmed in the computed styles above). The `default` arm only catches non-standard values like the bare attribute.

Two small thoughts: I'd move the `default` arm to the end of the `match` (valid where it is, but the end is where the eye expects it), and `field-sizing-content` / `resize-x` are Tailwind v4 utilities, which is fine since Flux already requires v4.

Nice cleanup — this also means user-supplied classes can actually override resize behavior now, which the old inline style silently blocked.

troccoli-sgr:
Awesome! Thanks guys.

rspahni:
Actually, rows="auto" (in combination mit resize="none") is no longer working in my app after this change. I get a textarea two lines high and not adjusting to the content. Happy to submit an issue if that's of any value. For now, I have overridden <flux:textarea> with its 2.15.0 predecessor and everything works again.

joshhanley:
@rspahni can you open an issue and we’ll have a look into it? Thanks!

rspahni:
> @rspahni can you open an issue and we’ll have a look into it? Thanks!

Done. Thx in advance for taking a look.

PR #2630, opened 2026-05-22
Refactor suffix rendering logic in `flux:menu.item` component

ghabriel25:
Just a tiny refactor for `flux:menu.item`

## Before

image

## After

image

## File Changed
- `stubs/resources/views/flux/menu/item.blade.php`

### DIFF
--- a/stubs/resources/views/flux/menu/item.blade.php
+++ b/stubs/resources/views/flux/menu/item.blade.php
@@ -61,14 +61,12 @@ $suffixClasses = Flux::classes()
 
     {{ $slot }}
 
-    <?php if ($suffix): ?>
-        <?php if (is_string($suffix)): ?>
-            <div class="{{ $suffixClasses }}">
-                {{ $suffix }}
-            </div>
-        <?php else: ?>
+    <?php if (is_string($suffix) && $suffix !== ''): ?>
+        <div class="{{ $suffixClasses }}">
             {{ $suffix }}
-        <?php endif; ?>
+        </div>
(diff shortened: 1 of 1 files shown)


joshhanley:
@ghabriel25 thanks for the PR! Personally, I prefer the before as I find it easier to read 😅 mainly because I know nothing in that block will happen if there is no suffix. But let's see what Caleb thinks

ghabriel25:
@joshhanley It's okay. This refactor just following how other components handle conditional slot/text

ganyicz:
A more compelling argument here would be that the component renders more correctly now.

Before, the truthiness check would disregard the suffix if the value is `0` string.

SCR-20260528-inai

ghabriel25:
@ganyicz Thanks!

calebporzio (commented):
Thanks @ghabriel25! I ran this through a real render comparison (fresh Laravel app, Flux as a path repo, DOM-diffed `menu.item` output on `main` vs this branch) across: no suffix, string suffix, `kbd`, and a suffix slot. Output is byte-identical in every case except one:

- `suffix="0"` — previously swallowed by the truthiness check, now renders. As @ganyicz pointed out, that's a correctness win, not a regression.

I also like that this now mirrors the `is_string($x) && $x !== ''` shape the `$icon` and `$iconTrailing` blocks in this same file already use, so the file reads consistently top to bottom.

I hear Josh on the readability of the original nesting, but consistency with the surrounding blocks plus the `"0"` fix tips it for me. Appreciate the tidy-up!

PR #2631, opened 2026-05-22
Replace blade directive with php syntax in `flux:pagination` component

ghabriel25:
Continuation of https://github.com/livewire/flux/pull/2596

## File Changed
- `stubs/resources/views/flux/pagination.blade.php`

### DIFF
--- a/stubs/resources/views/flux/pagination.blade.php
+++ b/stubs/resources/views/flux/pagination.blade.php
@@ -22,123 +22,123 @@ $scrollIntoViewJsSnippet = ($scrollTo !== null && $scrollTo !== false)
     : '';
 @endphp
 
-@if ($simple)
+<?php if ($simple): ?>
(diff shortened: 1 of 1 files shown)


joshhanley:
@ghabriel25 thanks for the PR! I actually think this file should use the Blade directives.

To give you some context, the reason why we opt for `<?php if(): ?>` a lot of the time in Flux is because once set, the conditionals generally don't change. So by using `<?php if(): ?>`  instead of `@if()` Livewire's morph markers aren't injected.

But in the case of pagination, it's a dynamic component that regularly changes based on the users interactions, because of that it requires the morph markers to ensure there aren't any morph issues.

Based on that, I'm going to close this. Thanks for your efforts on this!

ghabriel25:
@joshhanley I see. No problem, just venting my boredome 😆 . Thanks for the information!

joshhanley:
@ghabriel25 all good! 😁

PR #2649, opened 2026-06-09
Fix collapsed sidebar group keyboard navigation

### LINKED ISSUE #2643: Collapsible sidebar bugs: keyboard navigation, logo/icon toggle, focus
### Flux version

v2.14.1

### Livewire version

v4.3.0

### Tailwind version

v4.3.0

Chrome on Windows

### What is the problem?

There are a few bugs with the [collapsible sidebar](https://fluxui.dev/demo/sidebar-collapsible) that I've noticed. I've recorded a couple videos to demonstrate these issues.

1. Keyboard navigation doesn't work for collapsible link groups when the sidebar is collapsed, such as the `Favorites` in the menu here. I can expand the list, but I cannot focus the 3 items inside it. I've tried pressing space, tabbing, arrow keys.
2. The logo that toggles the sidebar switches between logo/expand icon inconsistently. Try this: collapse the sidebar, start with mouse outside, mouse over `Favorites` and move upwards over the other icons. The toggle sidebar icon will change to the logo again. If you move mouse back down to `Favorites` again, it shows the toggle sidebar icon again. If you move out of the sidebar, and then move into an empty space of the sidebar first, the logo disappears until you leave the sidebar.
3. Selecting an input outside the sidebar, and then hover the collapsed sidebar items causes the input to flash focus. Not sure if this is a necessary approach, but it can be a bit distracting, especially since there's a small gap between menu items.

It's mostly testable here: [collapsible sidebar](https://fluxui.dev/demo/sidebar-collapsible)

Some videos:

https://github.com/user-attachments/assets/bc496481-977e-4222-b9ed-efeb63016a9e

https://github.com/user-attachments/assets/618efa72-fd15-49f1-af9c-4bfa893f1662

1. I expect tabbing/arrow keys to work. I'm not an expert on accessibility, but it should work with keyboard navigation.
2. It should be consistent. Either show the logo when hovering the sidebar, or show the icon. I'm not a big fan of the logo/icon toggle in the first place, or the fact that you can click an empty space to expand the sidebar. I have missclicked empty spaces a few times, and accidentally expanded it, and then you cannot collapse it with the empty space click again. I don't think it's very intuitive the way it is now.
3. Not sure what the best solution is or if the current approach is necessary, but it felt wrong when I experienced it.

joshhanley:
This has a corresponding Flux Pro PR https://github.com/livewire/flux-pro/pull/506

# The Scenario

Currently, if you have a collapsed sidebar that contains an expanded sidebar group, you cannot focus the group trigger, open the dropdown, and then reliably navigate inside the dropdown using the keyboard.

Pressing Space or Enter opens the dropdown, but focus does not move to the first item. That leaves keyboard users without a working path into the dropdown links. When focus does land on an item, it can also show a focus ring instead of matching the active background treatment used by standard Flux dropdown menu items.

CleanShot 2026-06-09 at 19 37 47

```blade
<flux:sidebar collapsible>
    <flux:sidebar.nav>
        <flux:sidebar.group expandable expanded icon="star" heading="Favorites">
            <flux:sidebar.item href="#marketing-site">Marketing site</flux:sidebar.item>
            <flux:sidebar.item href="#android-app">Android app</flux:sidebar.item>
            <flux:sidebar.item href="#brand-guidelines">Brand guidelines</flux:sidebar.item>
        </flux:sidebar.group>
    </flux:sidebar.nav>
</flux:sidebar>
```

# The Problem

Collapsed sidebar group dropdowns render sidebar items inside a menu. `ui-menu` already focuses the first item when opened with ArrowDown, but collapsed sidebar group dropdowns also need that first-item focus when opened with Space or Enter.

The sidebar item styles also handled hover state in this dropdown context, but did not style `data-active` or suppress the item focus outline. That made keyboard-focused dropdown items visually differ from standard Flux menu items.

# The Solution

Keep the existing ArrowDown menu behaviour unchanged, and add a scoped Space/Enter path only when the dropdown has `data-flux-sidebar-group-dropdown`. When that scoped path opens the menu, the first menu item receives focus so arrow-key navigation works from inside the dropdown.

Also style `data-active` for sidebar items inside collapsed group dropdowns and hide the per-item focus outline in that dropdown context, matching normal menu active-state styling.

(dropdown menu items do have a background colour, just not shown in the compressed gif)
CleanShot 2026-06-09 at 19 34 02

Fixes livewire/flux#2643

### DIFF
--- a/stubs/resources/views/flux/sidebar/item.blade.php
+++ b/stubs/resources/views/flux/sidebar/item.blade.php
@@ -25,7 +25,8 @@ $tooltip ??= $slot->isNotEmpty() ? (string) $slot : null;
 // Size-up icons in square/icon-only buttons...
 $iconClasses = Flux::classes('size-4')
     ->add('in-data-flux-sidebar-group-dropdown:text-zinc-400! dark:in-data-flux-sidebar-group-dropdown:text-white/80!')
-    ->add('[[data-flux-sidebar-item]:hover_&]:text-current!');
+    ->add('[[data-flux-sidebar-item]:hover_&]:text-current!')
+    ->add('[[data-flux-sidebar-item][data-active]_&]:text-current!');
 
 $classes = Flux::classes()
     ->add('h-8 in-data-flux-sidebar-on-mobile:h-10 relative flex items-center gap-3 rounded-lg')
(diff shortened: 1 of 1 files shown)


calebporzio (approved):
Verified against a real collapsible-sidebar repro (Playwright, testbench harness): on main, Space on a collapsed group opens the dropdown but never focuses the first item; with this + flux-pro#506, Space/Enter focus the first item and arrow keys navigate. Covers sub-bug 1 of #2643.

PR #2657, opened 2026-06-16
Add configurable card indicator positions

joelstein:
# The scenario

Card-style radio and checkbox controls often place their indicator before the label. Both components should expose the same positioning API so an interface can use one convention consistently.

# The problem

Radio cards could only render their indicator at the end. Checkbox cards also fixed the indicator at the end, despite already accepting an `indicator` prop, so neither per-card nor group-level positioning was available consistently.

The 18px indicators were also top-aligned inside a 20px heading line box, placing their visual center 1px above the label.

# The solution

Both `radio` and `checkbox` card variants now use the same `indicator` semantics:

- `start` or `left` renders at logical start.
- `end`, `right`, `true`, or another truthy value renders at logical end.
- `false` hides the indicator.
- The position can be set on an individual card or inherited from its group.

`left` and `right` are compatibility aliases for logical `start` and `end`, so the layout follows RTL direction correctly. The existing default remains end-aligned.

Both indicators also receive a 1px top offset, aligning their center exactly with the label's line box.

# Validation

- Exercised radio and checkbox start/end positions in a real browser.
- Covered group inheritance and per-card overrides.
- Covered `left`, `right`, arbitrary truthy values, and `false`.
- Covered checked and unchecked interactions.
- Covered RTL logical direction.
- Measured indicator-to-heading center delta at `0px` for both controls.
- Browser console remained clean.

[Before/after screenshots and measurement details](https://github.com/livewire/flux/pull/2657#issuecomment-5292984089)

Radio card indicator positions

Fixes livewire/flux#2654.

### DIFF
--- a/stubs/resources/views/flux/checkbox/group/variants/cards.blade.php
+++ b/stubs/resources/views/flux/checkbox/group/variants/cards.blade.php
@@ -11,6 +11,7 @@
     'variant' => null,
     'size' => null,
     'name' => null,
+    'indicator' => true,
 ])
 
 @php
--- a/stubs/resources/views/flux/checkbox/variants/cards.blade.php
+++ b/stubs/resources/views/flux/checkbox/variants/cards.blade.php
@@ -14,6 +14,12 @@
 ])
 
 @php
+// Normalize the "indicator" prop into the same position semantics as radio cards...
+$indicatorPosition = $indicator ? 'end' : false;
+
+if ($indicator === 'start' || $indicator === 'left') {
+    $indicatorPosition = 'start';
+}
 
 $iconClasses = Flux::classes()
     ->add('inline-block mt-0.5 text-zinc-400 [ui-checkbox[data-checked]_&]:text-zinc-800 dark:[ui-checkbox[data-checked]_&]:text-white')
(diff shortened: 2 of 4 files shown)


ghabriel25:
Hi @joelstein Have you try this with Blaze enabled?

joelstein:
@ghabriel25 Yes, I tested it with Blaze enabled, and it works correctly.

joshhanley (approved):
@joelstein thanks for the PR! There was one small issue where `$indicator` was no longer being treated as truthy, like the original implemention had. So I've updated it to support that. Otherwise, all looks good to me.

calebporzio:
Reviewed — `indicator="start"` (with `left` as an alias) matches how `flux:switch` handles its `align` prop, and since the indicator is placed by DOM order it flips correctly in RTL. Josh's truthiness fix covers the back-compat cases. Pushed a one-word spelling fix (behaviour → behavior) to match the codebase. Follow-up for us: mirror this on `checkbox.cards` so the two card variants stay in sync, and document the new accepted values.

calebporzio:
Pushed `916fa8f` to make the vertical alignment geometrically consistent across both card controls.

The 18px indicator was top-aligned against the heading's 20px line box, placing its center 1px above the label. The update applies the same `mt-px` correction to:

- radio cards with the indicator at `start`
- radio cards with the indicator at `end`
- checkbox cards

Measured indicator-to-heading center delta:

- before: `-1px`
- after: `0px`

I also exercised both radio positions and checked/unchecked checkbox states in the browser; state changes work and the console is clean.

**Before**

alignment-before

**After**

alignment-after

calebporzio:
Follow-up: checkbox card **positioning** was not present—I had initially only mirrored the 1px vertical-centering correction.

Pushed `7304ccb` to give checkbox cards the full radio-card indicator API:

- `start` / `left` → logical start
- `end` / `right` / arbitrary truthy values → logical end
- `false` → no indicator
- group inheritance and per-card overrides
- correct logical placement in RTL

The PR title and description now cover both radio and checkbox cards. I exercised every state and alias in the browser, including checked/unchecked interaction; all indicator-to-heading center deltas remain exactly `0px`, and the console is clean.

**Final parity matrix**

checkbox-position-parity

PR #2665, opened 2026-06-24
Fix field spacing when only using description

### LINKED ISSUE #2664: Components without `label` are missing description padding
### Flux version

v2.13.2

### Livewire version

v4.2.4

### Tailwind version

v4.2.2

Chrome on MacOS

### What is the problem?

I noticed that the padding between the description and input field is missing when the `label` attribute is omitted.
Not sure if it also applies to other components.

The padding is **correct** when using `description:trailing` instead of `description`.

```blade
{{--  Correct padding --}}
<flux:editor description="Editor description" label="label" />
<flux:input description="Input description" label="Label" />

{{--  Incorrect padding --}}
<flux:editor description="Editor description" />
<flux:input description="Input description" />
```

Image

I would expect the same padding between the description and the input when the `label` attribute is missing, as when the `label` attribute is given or `description:trailing` is used.

joshhanley:
# The Scenario

When rendering Flux input fields, a description without a label sits directly against the input. The spacing is correct for a label-only field, and also for a field that has both a label and description.

CleanShot 2026-06-25 at 09 47 34@2x

```blade
<div class="space-y-8">
    <div>
        <flux:heading size="lg" level="2" class="mb-4">Label</flux:heading>

        <flux:input label="Input label" placeholder="Input value" />
    </div>

    <div>
        <flux:heading size="lg" level="2" class="mb-4">Label + description</flux:heading>

        <flux:input label="Input label" description="Input description" placeholder="Input value" />
    </div>

    <div>
        <flux:heading size="lg" level="2" class="mb-4">Description</flux:heading>

        <flux:input description="Input description" placeholder="Input value" />
    </div>
</div>
```

# The Problem

The field spacing selector handles `label+description` but not description on it's down. As such the description has no bottom margin when there is no label.

# The Solution

Add a field spacing selector for descriptions that are the first child and have following content. This gives description-only fields the same `mb-3` gap used between label descriptions and controls.

```php
'[&>[data-flux-description]:first-child:has(+*)]:mb-3',
```

The selector is scoped to that leading-description case, so existing label + description spacing remains unchanged.

CleanShot 2026-06-25 at 09 50 30@2x

Fixes #2664

### DIFF
--- a/stubs/resources/views/flux/field.blade.php
+++ b/stubs/resources/views/flux/field.blade.php
@@ -33,6 +33,7 @@ $classes = Flux::classes()
         default => [ // Adjust spacing around description...
             '[&>[data-flux-label]+[data-flux-description]]:mt-0',
             '[&>[data-flux-label]+[data-flux-description]]:mb-3',
+            '[&>[data-flux-description]:first-child:has(+*)]:mb-3',
             '[&>*:not([data-flux-label])+[data-flux-description]]:mt-3',
         ],
         'bare' => '',


calebporzio (approved):
Verified this against a fresh Laravel app with Flux as a path repo, measuring computed styles before/after:

| Case | main | this PR |
| --- | --- | --- |
| Description-only field (description `margin-bottom`) | `0px` (flush against input) | `12px` |
| Label + description | `12px` | `12px` (unchanged) |
| `description:trailing` below control | `mt: 12px` | `mt: 12px` (unchanged) |

The `:first-child:has(+*)` scoping is right — it only fires for a leading description with content after it, so a lone trailing description or label+description combos are untouched. Screenshot comparison confirms the description-only case now matches the label+description gap and everything else renders pixel-identical.

Thanks Josh!

PR #2695, opened 2026-07-21
Fix inset and padding on buttons with icons

ganyicz:
<details>
<summary>
Original PR
</summary>

# The scenario

When using a button with an icon, the padding is unbalanced

_The lines represent the default padding without an icon_

SCR-20260721-nwyy

# The problem

At this size, Heroicons come with inner padding of ~1px, pushing the icon ~2px from the edge (see 2nd and 3rd row)

We compensate for this on the default size, changing `ps-4/pe-4` (16px) to `ps-3/pe-3` (12px) when an icon is present (see 1st row)

The problem is that this is inconsistent:

1. We overcompensate by 2px on the default size 
2. We do not compensate for this on the smaller sizes

# The solution

1. Pull back the adjustment by 2px on the default size
2. Add the adjustment to the smaller sizes

SCR-20260721-nxfy

SCR-20260721-nnva

---

UPDATE:

I realized the original padding may have been adjusted for chevron which seems to have double the inner padding than the rest of the icons. With the above adjustments, the chevron is now pushed inside from the edges.

SCR-20260721-scjp-2
</details>

Revision:

# The scenario

Horizontal inset on a default size button doesn't correctly negate padding when an icon is present.

Frame@2x (9)

# The problem

When an icon is present, we intentionally reduce the padding on the side with the icon. This is to compensate for the padding inside the heroicon svg. Without this adjustment, the padding on the side with the icon would look bigger than on the side without it.

```php
($iconLeading ? 'ps-3' : 'ps-4') . ' ' . ($iconTrailing ? 'pe-3' : 'pe-4')
```

The problem is that inset doesn't account for this.

Additionally, the smaller sizes do not have this adjustment, which is why this bug only appears on the default size:

guidelines

# The solution

1. Add icon adjustment to smaller sizes
2. Make inset icon-aware on all sizes

guidelines-2

Frame@2x (10)

## Update

3. Reduce the gap on the `xs` size from `gap-2` to `gap-1`

SCR-20260724-okfv

### DIFF
--- a/stubs/resources/views/flux/button/index.blade.php
+++ b/stubs/resources/views/flux/button/index.blade.php
@@ -62,7 +62,7 @@ if ($loading && $type !== 'submit' && ! $isJsMethod) {
 }
 
 $classes = Flux::classes()
-    ->add('relative items-center font-medium justify-center gap-2 whitespace-nowrap')
+    ->add('relative items-center font-medium justify-center whitespace-nowrap')
     ->add('disabled:opacity-75 dark:disabled:opacity-75 disabled:cursor-default disabled:pointer-events-none')
     ->add(match ($align) {
         'start' => 'justify-start',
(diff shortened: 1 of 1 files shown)


calebporzio:
beauty!

PR #2726, opened 2026-08-10
Add toggle component

calebporzio:
## What changed

- add a free-tier `flux:toggle` component built on `ui-switch` semantics
- support `wire:model`, checked and disabled states, outline/solid icon transitions, text labels, tooltips, sizes, variants, and active icon colors
- use the outline button treatment by default while keeping the control surface neutral when toggled
- add Toggle to the Flux publish command
- allow the accent wrapper to inherit the application accent when no explicit color is supplied

## Why

Building a compact button-shaped binary control currently requires users to manually combine switch behavior, icon swapping, state styling, and accessibility. This component provides that treatment while reusing the existing switch behavior and form integration.

## Validation

- companion Playwright coverage exercises switch semantics, keyboard interaction, icon transitions, colors, variants, sizes, and disabled states
- 10 Toggle browser tests pass
- Flux docs production build passes

## Companion PRs

- livewire/flux-pro#541
- livewire/flux-docs#185

### DIFF
--- a/src/Console/PublishCommand.php
+++ b/src/Console/PublishCommand.php
@@ -49,6 +49,7 @@ class PublishCommand extends Command
             'Switch' => ['switch'],
             'Textarea' => ['textarea'],
             'Toast' => ['toast'],
+            'Toggle' => ['toggle'],
             'Tooltip' => ['tooltip'],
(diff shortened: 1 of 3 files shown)


utsavsomaiya:
Thanks 🥳

PR #2736, opened 2026-08-14
Add inset and fix padding in toggle

ganyicz:
# The scenario

The toggle doesn't support inset and padding / gap don't match the button

SCR-20260814-luiu

# The problem

Inset was never implemented and the padding / gap was likely based on older version of button before https://github.com/livewire/flux/pull/2695 was merged

# The solution

Adjust padding and gap to match the button and add support for inset

SCR-20260814-lvjh

SCR-20260814-mcqq


calebporzio:
sry, resolve conflicts then can merge tyty

PR #2738, opened 2026-08-14
Fix toggle interactions

ganyicz:
# The scenario

The toggle shows selection cursor and allows text selection, which is inconsistent with button

Additionally, clicking the toggle repeatedly causes zoom on mobile

https://github.com/user-attachments/assets/b2ac43b2-045c-437e-8f91-006d4d47bc88

# The problem

HTML button displays the default cursor and prevents selection / zoom by default.

The toggle is composed of `ui-switch` which does not have these properties.

# The solution

Add `select-none` and `touch-manipulation` classes to mimic button behavior.


calebporzio:
amazing. thanks

PR #2745, opened 2026-08-16
Add `-webkit-user-select: none`

ganyicz:
# The problem

Safari doesn't support `user-select: none` without a `-webkit` prefix.

In https://github.com/livewire/flux/pull/2738 we had to add `select-none` class to force the default cursor on toggle, despite the `ui-switch` already applying `user-select: none` in the default css rules. The reason was that the bug was limited to Safari. Because the `select-none` tailwind class applies the rule with the webkit prefix, this fixed the issue. However the fix really belongs to the css file.

# The solution

1. Add `-webkit-user-select: none` to all default CSS rules where we have `user-select: none`
2. Remove `select-none` from toggle

### DIFF
--- a/dist/flux.css
+++ b/dist/flux.css
@@ -155,14 +155,14 @@ Custom element display rules...
     ui-empty { display: block; cursor: default; }
     ui-pillbox { display: block; }
     ui-checkbox-group { display: block; }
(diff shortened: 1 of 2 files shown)


calebporzio:
@copilot resolve the merge conflicts in this pull request

PR #2749, opened 2026-08-17
Add soft card variant

calebporzio:
## Summary

- add a `soft` Card variant for a lower-emphasis surface
- preserve `outline` as the default Card treatment
- use translucent neutral overlays so the surface adapts to light, dark, and tinted backdrops

## Why

An opaque `zinc-50/100` treatment looks right on a white page, but can disappear when it is placed on the same zinc surface and can become visually inverted on darker zinc backgrounds.

The `soft` variant uses translucent black overlays in light mode and white overlays in dark mode. On the base background, these are tuned to visually match the original opaque treatment; on other backgrounds, they retain the backdrop hue while maintaining a consistent surface relationship.

## Screenshots

| Light | Dark |
| --- | --- |
| Outline, opaque reference, and soft cards across light backdrops | Outline, opaque reference, and soft cards across dark backdrops |

## Validation

- `composer validate --no-check-publish`
- interactively verified `outline`, the opaque reference treatment, and `soft` across white/zinc/blue backdrops in light and dark mode
- compared rendered screenshot pixels on the base backdrop: exact light fill/border match, exact dark border match, and a one-value dark fill difference

### DIFF
--- a/stubs/resources/views/flux/card/index.blade.php
+++ b/stubs/resources/views/flux/card/index.blade.php
@@ -1,13 +1,17 @@
 @blaze(fold: true)
 
 @props([
+    'variant' => 'outline',
     'size' => null,
 ])
 
 @php
 $classes = Flux::classes()
-    ->add('[:where(&)]:bg-white dark:[:where(&)]:bg-white/10')
-    ->add('border border-zinc-200 dark:border-white/10')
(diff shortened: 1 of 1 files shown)


calebporzio:
@copilot resolve the merge conflicts in this pull request

Issue #2608, opened 2026-05-11
Custom icon for `file-upload.dropzone`


lode:
### Flux version

v2.14.1

### Livewire version

v4.2.4

### Tailwind version

v4.1.17

### Browser and Operating System

Firefox on Ubuntu

### What is the problem?

Passing a custom icon name from Lucide to `<flux:file-upload.dropzone>` fails because that component forces `solid` icon variants, which Lucide doesn't have.

### Code snippets to replicate the problem

```bash
artisan flux:icon file-plus
```

```blade
<flux:file-upload.dropzone icon="file-plus">
```

### Screenshots/ screen recordings of the problem

This throws a `The "solid" variant is not supported in Lucide.` error.

### How do you expect it to work?

Like all other Flux components which does support Lucide icons.

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be closed if I haven't met the criteria above.

joshhanley:
@lode thanks for reporting! I've submitted a PR with a fix.

joshhanley:
Fix has been merged and will be in the next release.

PR #2763, opened 2026-08-21
Allow overriding the toggle icon variant

### LINKED ISSUE #2762: flux:toggle forces the solid icon variant, breaking Lucide icons
### Flux version

v2.17.0 (flux-pro 2.17.0)

### Livewire version

v4.4.1

### Tailwind version

v4.3.3

Chrome on macOS 15 (Darwin 25.5.0)

### What is the problem?

Passing a Lucide icon to `<flux:toggle>` fails because the component forces the `solid` variant for the "on" icon, which Lucide doesn't have.

`toggle.blade.php` renders the two icons like this:

```blade
<flux:icon :icon="$onIcon ?? $icon" variant="solid" :class="$iconClasses->add('hidden group-data-checked:block')" />
<flux:icon :icon="$offIcon ?? $onIcon ?? $icon" variant="outline" :class="$iconClasses->add('group-data-checked:hidden')" />
```

The `variant="solid"` on the first line is hardcoded, so any Lucide icon passed as `icon` or `on:icon` throws.

This is the same root cause as #2608 (`flux:file-upload.dropzone`). That one was fixed by adding an overridable `iconVariant` prop defaulting to `solid`. `flux:toggle` has no equivalent: it doesn't declare an `iconVariant` prop and doesn't pluck `icon:variant`, so `icon:variant="outline"` doesn't help either. It looks like the dropzone fix simply didn't reach `toggle`.

Heroicons work fine, and the same Lucide icon works in every other component I tried (`flux:button`, `flux:badge`), so this is specific to `flux:toggle`.

```bash
php artisan flux:icon shield-check
```

```blade
{{-- throws --}}
<flux:toggle icon="shield-check" />

{{-- also throws --}}
<flux:toggle on:icon="shrink" off:icon="expand" />

{{-- also throws: no escape hatch via icon:variant --}}
<flux:toggle icon="shield-check" icon:variant="outline" />

{{-- works: heroicon --}}
<flux:toggle icon="moon" />

{{-- works: same Lucide icon in other components --}}
<flux:button icon="shield-check" />
<flux:badge icon="shield-check">x</flux:badge>
```

Rendering any of the first three throws:

```
The "solid" variant is not supported in Lucide.
```

Rendering each snippet through `Blade::render()` gives:

```
THROW | toggle + lucide (icon=)     -> The "solid" variant is not supported in Lucide.
THROW | toggle + lucide (on:icon=)  -> The "solid" variant is not supported in Lucide.
THROW | toggle + icon:variant="outline" -> The "solid" variant is not supported in Lucide.
OK    | toggle + heroicon (icon=)
OK    | button + lucide (control)
OK    | badge + lucide (control)
```

The exception comes from the stub `flux:icon` generates for Lucide icons, which is thrown for any `solid` variant:

```php
// resources/views/flux/icon/shield-check.blade.php (generated by flux:icon)
if ($variant === 'solid') {
    throw new \Exception('The "solid" variant is not supported in Lucide.');
}
```

`<flux:toggle>` should accept Lucide icons like the other Flux components do.

Applying the same fix as #2608 would do it: declare an `iconVariant` prop (defaulting to `solid` so nothing changes for Heroicons) and honour `icon:variant`, so Lucide users can opt into `outline`. Falling back to `outline` automatically when the icon has no solid variant would be even nicer, since the checked/unchecked distinction is already carried by the background and colour.

cpswsg:
# The scenario

Passing a Lucide icon to `<flux:toggle>` throws `The "solid" variant is not supported in Lucide.`

```bash
php artisan flux:icon shrink expand
```

```blade
<flux:toggle on:icon="shrink" off:icon="expand" />
```

Heroicons work, and the same Lucide icon renders fine in `flux:button` and `flux:badge`, so this is specific to `flux:toggle`.

# The problem

`stubs/resources/views/flux/toggle.blade.php` hardcodes the `solid` variant for the checked icon:

```blade
<flux:icon :icon="$onIcon ?? $icon" variant="solid" :class="$iconClasses->add('hidden group-data-checked:block')" />
<flux:icon :icon="$offIcon ?? $onIcon ?? $icon" variant="outline" :class="$iconClasses->add('group-data-checked:hidden')" />
```

The stub `flux:icon` generates for Lucide icons throws on any `solid` variant:

```php
if ($variant === 'solid') {
    throw new \Exception('The "solid" variant is not supported in Lucide.');
}
```

`flux:toggle` neither declares an `iconVariant` prop nor plucks `icon:variant`, so there is no way to opt out.

This is the same root cause as #2608 for `flux:file-upload.dropzone`, which was fixed by adding an overridable `iconVariant` prop. That fix never reached `toggle`.

# The solution

Mirror the dropzone fix: add an `iconVariant` prop defaulting to `solid`, overridable via `icon:variant`, and use it for the checked icon.

```blade
@blaze(fold: true, unsafe: [
    'icon:class', 'icon:variant', 'on:icon', 'off:icon', 'on:label', 'off:label',
    ...
])

@php $iconVariant ??= $attributes->pluck('icon:variant'); @endphp

@props([
    'icon' => null,
    'iconVariant' => 'solid',
    ...
])

<flux:icon :icon="$onIcon ?? $icon" :variant="$iconVariant" ... />
```

Lucide icons then work with:

```blade
<flux:toggle on:icon="shrink" off:icon="expand" icon:variant="outline" />
```

The default is unchanged, so Heroicon toggles keep their solid checked icon.

I rendered these through `Blade::render()` on Flux v2.17.0 before and after the change:

| Case | Before | After |
| --- | --- | --- |
| `icon="expand" icon:variant="outline"` (Lucide) | throws | renders, outline |
| `on:icon="shrink" off:icon="expand" icon:variant="outline"` (Lucide) | throws | renders, outline |
| `icon="moon"` (Heroicon, default) | renders, solid | renders, solid |
| `on:icon="moon" off:icon="sun"` (Heroicon) | renders, solid | renders, solid |
| `<flux:toggle>Wi-Fi</flux:toggle>` (label only) | renders | renders |
| `wire:model="x" icon="moon"` | renders, solid | renders, solid |

## Known limitation

This is opt-in, exactly like the dropzone fix: `<flux:toggle icon="shield-check" />` still throws unless `icon:variant="outline"` is passed. Falling back automatically would require knowing which icons lack a solid variant, which the stub can't determine. Happy to take a different approach if you'd prefer the fallback be automatic.

When both icons end up outline, the checked/unchecked distinction is still carried by the background and text color the component already applies.

Fixes livewire/flux#2762

### DIFF
--- a/stubs/resources/views/flux/toggle.blade.php
+++ b/stubs/resources/views/flux/toggle.blade.php
@@ -1,10 +1,11 @@
 @blaze(fold: true, unsafe: [
-    'icon:class', 'on:icon', 'off:icon', 'on:label', 'off:label',
+    'icon:class', 'icon:variant', 'on:icon', 'off:icon', 'on:label', 'off:label',
     'tooltip:position', 'tooltip:kbd', 'tooltip',
 ])
 
 @php $onLabel ??= $attributes->pluck('on:label'); @endphp
 @php $offLabel ??= $attributes->pluck('off:label'); @endphp
+@php $iconVariant ??= $attributes->pluck('icon:variant'); @endphp
 @php $onIcon ??= $attributes->pluck('on:icon'); @endphp
 @php $offIcon ??= $attributes->pluck('off:icon'); @endphp
 
@@ -14,6 +15,7 @@
     'size' => 'base',
     'name' => null,
     'icon' => null,
+    'iconVariant' => 'solid',
     'label' => null,
     'color' => null,
     'inset' => null,
@@ -105,7 +107,7 @@ $classes = Flux::classes()
     <flux:with-tooltip :$attributes>
         <ui-switch {{ $attributes->class($classes) }} @if($showName) name="{{ $name }}" @endif @if($checked) checked data-checked @endif data-flux-control data-flux-toggle>
             <?php if ((is_string($icon) && $icon !== '') || $onIcon): ?>
-                <flux:icon :icon="$onIcon ?? $icon" variant="solid" :class="$iconClasses->add('hidden group-data-checked:block')" />
+                <flux:icon :icon="$onIcon ?? $icon" :variant="$iconVariant" :class="$iconClasses->add('hidden group-data-checked:block')" />
                 <flux:icon :icon="$offIcon ?? $onIcon ?? $icon" variant="outline" :class="$iconClasses->add('group-data-checked:hidden')" />
             <?php elseif ($icon): ?>
                 {{ $icon }}


calebporzio:
Thanks for the contribution! However i think instead I'm just going to recommend for lucide icons you pass the different icons you want into their relative slots.

i added docs for this that will be live today. thanks!

PR #2768, opened 2026-08-24
`flux:sidebar.item` defaults the tooltip to the slot, which may contain html

lode:
# The scenario

I'm using conditions in the slot of a `flux:sidebar.item`. When collapsing the sidebar the tooltip is shown. And when I don't set an explicit tooltip, it falls back to the slot. But since a tooltip is text-only it can't render html so well.

# The problem

```blade
@php($foo = 'bar')
<flux:sidebar.item icon="question-mark-circle" href="...">
    Support
    @if ($foo)
        {{ $foo }}
    @endif
</flux:sidebar.item>
```

With conditions Livewire actually outputs html comments, which end up in the tooltip.

image

But other html would also break this.

image

# The solution

I've been trying to avoid using conditions in this way, e.g. `{{ 'Support ' . ($foo ?? '') }}`, but that feels a bit annoying.

Stripping html sounds like a good addition which prevents stuff which we should anyway never render in a tooltip because it is at least ugly.

### DIFF
--- a/stubs/resources/views/flux/sidebar/item.blade.php
+++ b/stubs/resources/views/flux/sidebar/item.blade.php
@@ -20,7 +20,7 @@
 ])
 
 @php
-$tooltip ??= $slot->isNotEmpty() ? (string) $slot : null;
+$tooltip ??= $slot->isNotEmpty() ? strip_tags((string) $slot) : null;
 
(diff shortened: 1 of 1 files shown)


calebporzio:
perfect. thanks!

